PR #461, opened 2026-05-02
fix(hooks): honor user git-config in commands that fire hooks

### LINKED ISSUE #456: HooksConfig::default() ignores user settings across all commands
## Summary

Every command that fires hooks constructs a `HooksConfig::default()` instead of loading from git config via `load_hooks_config_global()` / `load_hooks_config()`. As a result, user-configured hook settings are silently ignored everywhere.

## Affected settings

From `keys::hooks::*`:
- `daft.hooks.enabled`
- `daft.hooks.defaultTrust`
- `daft.hooks.userDirectory`
- `daft.hooks.timeout` ← **most user-visible**: long-running cleanup hooks (docker teardown, db dump) silently time out at the 300s default even when the user configured a larger budget
- `daft.hooks.output.quiet`
- `daft.hooks.output.tailLines`

## Affected sites

```
$ grep -rn 'HooksConfig::default()' src/commands/
src/commands/clone.rs:554
src/commands/clone.rs:622
src/commands/clone.rs:862
src/commands/clone.rs:930
src/commands/clone.rs:1278
src/commands/clone.rs:1329
src/commands/flow_eject.rs:125
src/commands/sync.rs:486
src/commands/sync.rs:1299
src/commands/worktree_branch.rs:378
src/commands/worktree_branch.rs:472
```

`src/commands/repo/remove.rs` was the only command honoring hook config (added in PR #448 after review feedback). Every other command that runs hooks needs the same treatment.

## Root cause

`load_hooks_config_global()` and `load_hooks_config()` exist in `src/core/settings.rs:748,815` and have full population logic — they're just not called by any command. They were defined but never wired up.

## Proposed fix

Replace each `HooksConfig::default()` call site with `load_hooks_config_global()?` (or the per-repo `load_hooks_config()` if a repo context is available). Pre-clone callsites obviously need the global variant; post-clone ones can use the per-repo variant for finer-grained control.

For commands that branch on cwd-availability (e.g. `clone` at line 554 vs 1278), pick whichever loader doesn't fail when the cwd isn't a git repo — usually `_global`.

## Risk

Low. The loaders are infallible against missing keys (they only override defaults when present), and the only real failure mode is git-config probe failure which we already handle elsewhere via the same loaders.

## Testing

Add a unit/integration test that:
1. Sets `daft.hooks.timeout = 600` in global git config
2. Invokes a command that fires hooks (e.g. `daft worktree branch foo`)
3. Asserts the executed hook config has `timeout = 600`, not `300`

## Context

Surfaced by claude review on PR #448 ("Background hook coordinator" PR). Local fix landed there for `repo remove` only — codebase-wide propagation tracked here.

avihut:
## Summary

- Replace 21 `HooksConfig::default()` call sites across 11 command modules with `load_hooks_config[_global]()`. Every `daft.hooks.*` setting users wrote in their git config (`enabled`, `timeout`, `userDirectory`, `output.quiet`, ...) was previously ignored — the loaders existed but were never called outside `repo remove`.
- Pre-clone sites in `clone.rs` use `_global` (cwd may not be inside any repo). All in-repo commands use `load_hooks_config()` to pick up repo-local overrides too.
- Hoist the loader call out of `create_satellite_worktrees`'s per-branch loop and out of the `--all-branches` orchestrator's two construction sites — load once, clone per-iteration. Multi-branch clones no longer re-read global git config N times.

Fixes #456

## Test plan

- [x] `mise run fmt` clean
- [x] `mise run clippy` zero warnings
- [x] New regression test `tests/manual/scenarios/hooks/config-disabled-respected.yml` verified to **fail on master** (marker file created because `enabled=true` default ignored the user's `daft.hooks.enabled=false`) and **pass with the fix**
- [x] `clone/all-branches-hooks.yml` still passes (validates the loader-hoist refactor in `clone.rs`)
- [x] All other hook scenarios continue passing

### DIFF
--- a/src/commands/branch_delete.rs
+++ b/src/commands/branch_delete.rs
@@ -1,7 +1,7 @@
 use crate::{
     core::{worktree::branch_delete, CommandBridge},
     git::should_show_gitoxide_notice,
-    hooks::{HookExecutor, HooksConfig},
+    hooks::HookExecutor,
     is_git_repository,
     logging::init_logging,
     output::{CliOutput, Output, OutputConfig},
@@ -102,7 +102,7 @@ fn run_branch_delete(args: &Args, output: &mut dyn Output, settings: &DaftSettin
         prune_cd_target: settings.prune_cd_target,
     };
 
-    let hooks_config = HooksConfig::default();
+    let hooks_config = crate::core::settings::load_hooks_config()?;
     let executor = HookExecutor::new(hooks_config)?;
 
     if should_show_gitoxide_notice(settings.use_gitoxide) {
--- a/src/commands/checkout.rs
+++ b/src/commands/checkout.rs
@@ -11,7 +11,7 @@ use crate::{
     get_current_worktree_path, get_git_common_dir, get_project_root,
     git::{should_show_gitoxide_notice, GitCommand},
     hints::maybe_show_shell_hint,
-    hooks::{yaml_config_loader, HookExecutor, HooksConfig, TrustDatabase},
+    hooks::{yaml_config_loader, HookExecutor, TrustDatabase},
     is_git_repository,
     logging::init_logging,
     output::{CliOutput, Output, OutputConfig},
@@ -722,7 +722,7 @@ fn run_checkout(
         at_path: args.at.clone(),
     };
 
-    let hooks_config = HooksConfig::default();
+    let hooks_config = crate::core::settings::load_hooks_config()?;
     let executor = HookExecutor::new(hooks_config)?;
 
     if should_show_gitoxide_notice(settings.use_gitoxide) {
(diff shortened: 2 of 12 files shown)


avihut:
Thanks for the review @claude. Triage:

**Addressed (commit 9af5e1d):**
- **#1** & **#3** — added comments explaining the `_global` choice in `run_post_clone_hook`, `run_post_create_hook`, and `create_satellite_worktrees`. Local loader requires being inside a repo (`gix_repo()?` errors otherwise), so `_global` is the cwd-tolerant choice for paths that may run with cwd outside any repo. Repo-local overrides on a fresh clone are vanishingly rare — explicit tradeoff documented in code.

**Deliberately skipped:**
- **#2** (flow_adopt double-load) — your own assessment was "follow-up if hot, consistent with checkout.rs pre-existing pattern". Not in scope for #456.
- **#4** (sync.rs redundant `.clone()`) — pre-existing, not a regression.
- **#5** (test covers one site) — your assessment was that this satisfies CLAUDE.md's regression-test rule; all 21 sites use the identical loader call shape, so testing the loader once verifies the architecture.

PR #507, opened 2026-05-15
fix(tui): keep size column TOTAL summary intact when wider than data cells

### LINKED ISSUE #501: fix(tui): size column TOTAL summary truncates when total is wider than data cells
## Problem

The `daft list` TOTAL summary row truncates its size value when the formatted total string is wider than the data cells above it. In worktree-heavy repos with multi-GB worktrees, this looks like:

```
BRANCH       SIZE
feat/a       1.2G
feat/b       1.7G
main         3.4G
…
TOTAL        10.2     ← should be 10.2G, but the "G" is cut off
```

The data rows render fine because `1.2G` / `1.7G` are short. The TOTAL cell is wider (`10.2G`, `127.4G`, etc.) but the column's width has already been frozen to the widest data cell — so the summary value is silently truncated to fit.

This is the inverse of how the size column *should* behave: when a summary row is present, the column width must accommodate the wider of (max data cell, summary cell).

## Reproducer

```bash
cd /path/to/repo-with-many-large-worktrees
daft list --columns +size
# or
daft list --stat summary    # depending on flag wiring
```

In `src/commands/list.rs:605-628` the TOTAL row is built and appended to the existing `table` after the data rows. By this point the column-width policy has been decided based on data only.

## Suggested fix direction

`src/commands/list.rs` builds the table via `tabled` for the blocking renderer. The fix is one of:

1. **Pre-compute summary cell width and feed it into the column-width policy** before `table.with(Width::truncate(...))` is applied. `column_content_width` (in `src/output/tui/columns.rs`) already takes this approach for the TUI; the blocking-list path needs the equivalent.

2. **Append the TOTAL row via a separate `tabled::Builder` configured with the same column widths**, so the summary cell gets its own per-row width budget rather than being forced into the data-row schema.

Option 1 is cleaner and matches the TUI's logic. Option 2 is a smaller diff but introduces a divergence between data and summary rendering that future bugs would have to remember.

Verify the same fix applies to the live TUI path (`src/output/tui/render.rs` + `src/output/tui/columns.rs`) — likely already correct since `column_content_width` is invoked over all rows uniformly, but a regression test for the wide-TOTAL case in `list_live` is worth adding.

## Acceptance criteria

- [ ] `daft list --columns +size` on a repo with N≥3 multi-GB worktrees shows the full TOTAL value (no missing unit suffix or trailing digits).
- [ ] Regression test exercises the wide-TOTAL case (data cells short, summary cell long) for both blocking and live-TUI render paths.
- [ ] No layout regression on the narrow case where data cells are wider than the summary.

avihut:
## Summary

- The `daft list --columns +size` TOTAL summary cell silently truncated when wider than per-row cells (`10.2G` → `10.2`, losing the unit suffix).
- Two render paths, two distinct mechanisms: TUI sizes columns from data rows only; blocking shrinks the size column under `Width::truncate` + `Priority::max(true)` once Branch/Path hit their floor.
- Fix threads the formatted TOTAL width into `column_content_width` (TUI) and replaces `Priority::max(true)` with a custom `PriorityMaxExcept` Peaker that excludes the size column (blocking).

## Why not `MinWidth`

`Width::truncate` derives per-column floors from `EmptyRecords` (padding only), not from `MinWidth` settings — verified empirically with a standalone repro. `MinWidth`/`Width::increase` pads cells but doesn't pin a shrink floor. Excluding the column from the candidate set is the only surgical fix.

## Out of scope

The other fixed-width metric columns (`Hash`, `Changes`, `Base`, `Remote`, `Age`, `Owner`) have the same class of bug under `Priority::max(true)` in very narrow terminals — the TUI's `fit_widths_to_available` only shrinks `{Branch, Path, LastCommit}`. Noted in `PriorityMaxExcept`'s docstring; can be a follow-up.

## Test plan

- [x] `mise run fmt:check` — clean
- [x] `mise run clippy` — clean
- [x] `mise run test:unit` — 1713 tests pass
- [x] `mise run test:manual -- --ci tests/manual/scenarios/list/` — 53 scenarios / 167 steps pass
- [x] New `TestBackend` regression test fails pre-fix (renders `15.0`), passes post-fix (renders `15.0G`) — verified by temporarily reverting `column_content_width`'s `.max(extra_width)`
- [x] Standalone tabled repro confirms blocking-path fix at terminal widths 18, 20, 25, 30, 120
- [ ] Manual TTY check in a scratch repo with multi-GB worktrees (CI cannot exercise this — `terminal_size()` returns `None` under piped stdio)

Fixes #501

### DIFF
--- a/src/commands/list.rs
+++ b/src/commands/list.rs
@@ -27,7 +27,11 @@ use clap::Parser;
 use std::collections::HashSet;
 use tabled::{
     builder::Builder,
-    settings::{Padding, Style, Width, object::Columns, peaker::Priority},
+    settings::{
+        Padding, Style, Width,
+        object::Columns,
+        peaker::{Peaker, Priority},
+    },
 };
 use terminal_size::{Width as TermWidth, terminal_size};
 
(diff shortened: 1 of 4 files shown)


avihut:
Rebased onto current `origin/master` (now includes #506 — `8adcecc`). Force-pushed.

Re: review observations:

- **#1 (Vec::contains for excluded)**: agreed, no change.
- **#2 (blocking path omits pruned filter)**: this is actually not a discrepancy — the blocking renderer iterates `WorktreeInfo`, which has no `status` field; the pruned-filter only makes sense in the TUI where rows have a `WorktreeStatus`. Different inputs by design, not pre-existing inconsistency. Skipping.
- **#3 (`as u16` cast)**: micro-nit; the `format_human_size_has_no_internal_whitespace` test pins the contract that bounds the input. Skipping the `debug_assert!`.
- **#4 (`extra_width` single-valued)**: appropriately minimal; documented.
- **#5 (YAML scenario coverage gap)**: documented in the scenario itself, mirroring #497's pattern.

Local re-validation post-rebase: build green, `output::tui::columns::tests` (13) ✓, `render_table_size_column_total_fits_when_wider_than_data` ✓, `commands::list::tests` (10) ✓, pre-push hook (build-check + unit-tests) ✓.

PR #557, opened 2026-05-25
feat(test-runner): default to automatic; add --interactive (-i) for step-through

### LINKED ISSUE #554: feat(test-runner): default to automatic mode; add --interactive (-i) for step-through
## Motivation

The YAML scenario runner under `xtask/src/manual_test/` started life as a step-by-step interactive debugger and grew into the project's integration test suite — 581+ scenarios that get exercised on every PR by `mise run test:manual -- --ci`. That growth has inverted the user model:

- **What people actually do:** run the whole suite in automatic mode (`--ci`) — both locally and in CI. This is the everyday flow.
- **What the flag set assumes:** interactive step-through is the default; automatic is the opt-in (`--no-interactive` aliased as `--ci`).

The flag's name is now a vestigial artifact. Every routine invocation types `-- --ci` and the no-flag invocation does something almost no one wants (interactively step through 581 scenarios on a TTY).

Flip the default so the common case becomes implicit: automatic on a TTY by default, `--interactive` / `-i` to opt in to step-through. This makes the runner read as what it actually is — an integration test runner — rather than a debugger that grew a CI mode.

## Today vs proposed

| | Today | Proposed |
|---|---|---|
| Default on TTY (no flags) | Interactive step-through | Automatic (parallel, buffered, live progress region) |
| Force automatic | `--ci` / `--no-interactive` | (default, no flag) |
| Force interactive | (default on TTY) | `--interactive` / `-i` |
| Default on non-TTY | Automatic (unchanged) | Automatic (unchanged) |

Per [`feedback_no_pre_1_0_back_compat`](https://github.com/avihut/daft/issues/518): cut `--no-interactive` / `--ci` cleanly — no deprecated aliases, no shims. Pre-1.0 surface. Any external caller typing `--ci` will get a clap error pointing at the new shape, which is more discoverable than silent magic.

### Pure logic change — 2 files, ~5 lines

**`xtask/src/main.rs:368-369`** — invert the flag:
```rust
// today
#[arg(long, alias = "ci")]
no_interactive: bool,

// after
#[arg(long, short = 'i')]
interactive: bool,
```

**`xtask/src/manual_test/mod.rs:385`** — invert the derivation:
```rust
// today
let is_interactive = !no_interactive && std::io::stdin().is_terminal();

// after
let is_interactive = interactive;   // explicit opt-in; TTY detection no longer routes
```

Plus the error message at `mod.rs:394` (`"--jobs/--parallel is only supported in non-interactive mode (pass --ci or run from a non-TTY)"` → `"--jobs/--parallel is incompatible with --interactive / -i"`).

### Live callers to update (`s/ --ci//g`)

| File | What |
|---|---|
| `xtask/src/main.rs:1366` | Internal smoke spawn in `runner-output` test |
| `xtask/src/bench.rs:328` | Bench smoke invocation |
| `.github/workflows/test.yml:486` | CI workflow |
| `benches/scenarios/test_manual_scale.sh` | Bench script |
| `mise-tasks/bench/tests/integration/_bench_command` | mise task |
| `mise-tasks/bench/tests/manual/_default` | mise task |
| `mise-tasks/sandbox/_default` | mise task |
| `tests/manual/scenarios/runner-output/end-to-end.yml` | The runner-output smoke fixture itself |

### Docs

- `tests/README.md` — reframe the TTY/CI paragraph
- `xtask/src/manual_test/reporter/{CLAUDE,mod,pretty,tests}.rs` — update doc comments that reference the `--ci` flag (the non-TTY / automatic concepts remain correct)
- `docs/about/contributing.md` — any `--ci` reference

Don't touch the historical plan docs under `docs/superpowers/plans/` — those are immutable artifacts.

## One subtle decision

`--step N` and `--loop-count N` only make sense in interactive mode (single-scenario step debugging). Today they're honored under `run_serial` regardless. After the flip, two options:

1. **Auto-imply `-i` when `--step` is passed** (silent magic — `--step 3` just works)
2. **Error out** (explicit — `--step 3 requires --interactive / -i`)

Recommend **option 2** — matches the inversion's "explicit over implicit" spirit. If someone passes `--step` without realizing what it does, a clap error is more discoverable than silently switching modes.

## Out of scope

- Renaming the suite or the runner (`mise run test:manual` stays, `xtask manual-test` stays). The conversation that prompted this issue framed the runner as an "integration test runner," but renaming is separate churn and orthogonal to the default-flip.
- Extracting the runner into its own crate ([referenced post-launch goal in #509](https://github.com/avihut/daft/issues/509)).
- Anything about output formatting or progress rendering — that's #518's territory.

## What does NOT change

- All runner internals (`run_serial`, `run_parallel`, sandbox lifecycle, cancellation paths, progress sink, reporter) — the `is_interactive` boolean still carries the same meaning downstream, only its source inverts.
- `--setup-only` — still always routes to `run_serial` regardless of `-i`.
- TTY detection for the live progress region (`show_progress` in `run_parallel`) — independent from routing, stays as-is.
- `--jobs > 1` × interactive incompatibility check — still works, only the flag name in the error message changes.

## Acceptance criteria

- [ ] `mise run test:manual` (no flags) runs in automatic mode on a TTY, parallel scheduler engaged.
- [ ] `mise run test:manual -- -i <scenario>` runs in interactive step-through mode.
- [ ] `--ci` and `--no-interactive` are removed cleanly (clap rejects with a useful error pointing at `-i`).
- [ ] `--step N` without `-i` errors with a clap message naming `--interactive`.
- [ ] All in-tree callers (CI workflow, mise tasks, bench scripts, internal smoke fixtures) updated.
- [ ] `tests/README.md` paragraph reframed.
- [ ] No behavioral regression in CI — `mise run test:manual` (now flagless in the workflow) produces the same output the `--ci` form produced.

## Related

- **#509** umbrella — adds this as a new sub-task.
- **#518** (resolved by current branch `daft-518/feat/test-runner-output-improvements`) — the conversation that surfaced this default-inversion happened while finishing #518.
- **#510** parallel execution — relies on the automatic path; this issue makes that path the default route.

## Estimate

- Logic: ~10 lines across 2 files
- Caller updates: ~10 lines across 8 files (mostly removing `--ci`)
- Doc updates: ~30 lines across 4–5 files
- Total: ~50 lines, one commit, no architectural ripple

avihut:
## Summary

- Flip the YAML test runner's default: automatic (parallel-buffered) is now implicit; `--interactive` / `-i` is the explicit opt-in to step-through. Cuts `--ci` / `--no-interactive` cleanly per `feedback_no_pre_1_0_back_compat` — clap rejects the old shape with a useful error pointing at `-i`.
- Adds explicit validation: `--step` requires `-i` or `--setup-only`; `--loop-count` requires `-i`. The documented `--setup-only --step N` workflow (`docs/about/contributing.md:131`) is preserved because setup-only honors `step` as a "run up to step N" bound independent of mode.
- Strips `--ci` from every in-tree caller (CI workflow, mise tasks, bench scripts, internal smoke spawns, YAML fixture) and updates sandbox completions/help + reporter reproduce hint + design language + developer docs.

Runner internals (`run_serial`, `run_parallel`, sandbox lifecycle, cancellation, reporter) do not change. The `is_interactive` boolean still carries the same downstream meaning — only its source inverts.

Commits land as three logical units: core flip + reporter goldens; sandbox surfaces; developer docs.

Fixes #554

## Test plan

- [x] `mise run fmt:check` — passes
- [x] `mise run clippy` — passes with zero warnings
- [x] `mise run test:unit` — 1816 passed, 0 failed
- [x] `manual-test --ci` rejected by clap (`unexpected argument '--ci'`)
- [x] `manual-test --no-interactive` rejected by clap (`a similar argument exists: '--interactive'`)
- [x] `manual-test --step 1 checkout:basic` errors with `--step requires --interactive / -i (or --setup-only)`
- [x] `manual-test --loop-count 3 --step 1 checkout:basic` errors with `--loop-count requires --interactive / -i`
- [x] `manual-test --setup-only --step 1 checkout:basic` still works (regression guard for `contributing.md:131`)
- [x] `manual-test runner-output:end-to-end` smoke passes (runner-output fixture now exercises the new default)
- [x] Runtime reproduce hint emits `mise run test:manual -- <token>` (no flag)
- [x] `rg "--ci\b|--no-interactive\b|no_interactive"` outside `docs/superpowers/plans/` and `benches/baselines/` returns no hits
- [ ] CI matrix passes on PR

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -483,7 +483,7 @@ jobs:
         # No `${{ github.event.* }}` substitutions touch the run: body.
         run: |
           GIT_CONFIG_GLOBAL="${{ steps.config.outputs.config_file }}" \
-            DAFT_TESTING=1 ./target/release/xtask manual-test --ci --jobs "$(nproc)"
+            DAFT_TESTING=1 ./target/release/xtask manual-test --jobs "$(nproc)"
 
       - name: Test shell completions
         if: matrix.suite == 'bash'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,9 +58,9 @@ mise run dev                # Build + create symlinks (quick dev setup)
 mise run test               # Run all tests (unit + integration)
 mise run test:unit          # Rust unit tests only
 mise run test:integration   # Integration tests (bash + YAML, full matrix)
-mise run test:manual -- --ci           # YAML manual tests only (all scenarios)
-mise run test:manual -- --ci checkout  # YAML tests for one command
-mise run test:manual -- checkout:basic # Interactive mode for one scenario
+mise run test:manual                       # YAML manual tests (all scenarios, automatic)
+mise run test:manual checkout              # YAML tests for one command (automatic)
+mise run test:manual -- -i checkout:basic  # Step through one scenario interactively
 mise run clippy             # Lint (must pass with zero warnings)
 mise run fmt                # Auto-format code
 mise run fmt:check          # Verify formatting
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -310,7 +310,7 @@ dependencies = [
  "find-msvc-tools",
  "jobserver",
  "libc",
- "shlex 1.3.0",
+ "shlex",
 ]
 
 [[package]]
@@ -748,7 +748,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "serial_test",
- "shlex 2.0.1",
+ "shlex",
  "similar",
  "syntect",
  "tabled",
@@ -4198,12 +4198,6 @@ name = "shlex"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fda2ff0d084019ba4d7c6f371c95d8fd75ce3524c3cb8fb653a3023f6323e64"
-
-[[package]]
-name = "shlex"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8fadd59c855ef2080decdef8ff161eb6661b86933c9d82e5ba29dc602a55aba"
 
 [[package]]
 name = "signal-hook"
--- a/benches/scenarios/test_manual_scale.sh
+++ b/benches/scenarios/test_manual_scale.sh
@@ -77,7 +77,7 @@ hyperfine \
     --command-name "jobs={jobs}" \
     --export-json "$JSON_OUT" \
     --export-markdown "${MD_OUT}.phase1" \
-    "$REPO_ROOT/target/release/xtask manual-test --ci --jobs {jobs}"
+    "$REPO_ROOT/target/release/xtask manual-test --jobs {jobs}"
 HF_STATUS=$?
 set -e
 if [[ $HF_STATUS -ne 0 ]]; then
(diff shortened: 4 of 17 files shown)


avihut:
Thanks for the review — addressing the one flagged sanity check:

**`Cargo.lock` shlex consolidation — verified benign.**

`Cargo.toml` declares `shlex = "1.3"` (= `^1.3`, i.e. `>=1.3.0, <2.0.0`) and uses it at `src/core/worktree/exec/mod.rs:740,782` via `shlex::try_quote`. The Dependabot bump #553 (commit `9d868dbd`) **only changed `Cargo.lock`** — it never touched `Cargo.toml`. That left an internally inconsistent state where the lockfile pinned daft's `shlex` to `2.0.1` while the manifest constraint forbade `2.x`.

`cargo check` in this worktree re-resolved the lockfile back to the version Cargo.toml actually permits, which is the diff that landed in this PR. `cargo tree -p shlex --invert` confirms a single `shlex v1.3.0` consumed by both `daft` directly and the cc build-dep chain — no silent downgrade, no version-skew. The reviewer's concern is the right one to raise, but the resolution here was the lockfile catching up to Cargo.toml, not the other way around.

#553 should probably be revisited separately — if the intent was to actually move to `shlex 2.x`, Cargo.toml needs a corresponding `shlex = "2"` bump. As it stands now, that bump was a no-op.

The other two findings (cosmetic `let is_interactive = interactive;` and absent unit tests for the new validation `anyhow::bail!` paths) are intentional — the rename preserves the established downstream `is_interactive` naming convention, and the validation block is consistent with how the pre-existing `--loop-count requires --step` guard was handled (verified by the test plan rather than unit tests). Happy to add validation-block unit tests in a follow-up if you'd prefer them as a uniform policy.

PR #558, opened 2026-05-25
fix(hooks): support background jobs depending on foreground jobs

### LINKED ISSUE #556: fix(hooks): bg job depending on fg job silently fails to run
## Symptom

A background hook job that declares `needs: [<some-foreground-job>]` is reported as `running` by the parent (`⟳ 1 background job running — daft hooks jobs to manage`) but **never actually executes**: no log file, no SQLite `jobs` row, no observable side effects.

This is exactly the canonical pattern shown in `SKILL.md` lines 506–516:

```yaml
hooks:
  worktree-post-create:
    jobs:
      - name: install deps          # foreground
        run: pnpm install
      - name: warm build cache      # background
        run: cargo build
        background: true
        needs: [install deps]       # ← bg depends on fg
```

This repo's own `daft.yml` hits the same bug (`mise dev` is bg and `needs: [bun install]` which is fg).

## Root cause

1. `partition_foreground_background` (`src/hooks/yaml_executor/partition.rs:15`) splits jobs into FG and BG slices but returns `JobSpec`s with `needs:` unmodified. A BG job retains references to FG names.
2. `spawn_coordinator` (`src/coordinator/process.rs:1299`) hands the BG slice to a child process spawned with `Stdio::null()` on stdin/stdout/stderr, and the parent unconditionally prints `⟳ N background job(s) running` (`src/hooks/yaml_executor/mod.rs:450`) without checking the child's startup outcome.
3. In the coordinator, `run_all_with_cancel` (`src/coordinator/process.rs:148–154`) feeds the BG slice to `DagGraph::new`, which constructs `name_to_idx` from BG-only names. Any `needs:` entry pointing at a FG name is rejected with `DagError::MissingDependency` (`src/executor/dag.rs:49`).
4. The coordinator catches the error, prints `daft: coordinator error: …` to stderr, and exits 1 (`src/coordinator/process.rs:1437–1441`) — but stderr is `/dev/null`, so the message is swallowed. No `JobRow` is ever written because `upsert_job` happens inside `run_single_background_job`, which is never reached.

The failure mode is even codified as a passing test: `bg_missing_dep_in_needs_returns_error` at `src/coordinator/process.rs:2648`. The author of #463 was aware that dangling `needs:` would error, but didn't connect it to the fact that the partitioner produces exactly this state for every BG job that depends on a FG job.

## Why it went unnoticed

- The parent's `⟳ N background job(s) running` message is unconditional — there's no liveness check on the coordinator child.
- `Stdio::null()` on the coordinator's stderr swallows the startup error.
- The canonical SKILL.md example is the failing pattern, but the example job (`warm build cache`) has no obvious user-visible side effect when it silently doesn't run, so nobody notices.

## Git origin

Introduced by `d004c799` (PR #463, `fix(coordinator): honor needs: between background jobs`). The pre-#463 coordinator had an unconditional fan-out loop that silently ignored `needs:`, so dangling cross-partition references silently worked. #463 replaced the loop with `DagGraph`-based scheduling, which rejects them.

## Fix shape

Two complementary changes:

1. **Strip BG `needs:` of FG names in the partitioner.** FG-before-BG sequencing already guarantees the "happens-after" intent — the cross-partition dep is a no-op the partitioner should consume.
2. **Cascade FG failure into BG.** If a BG job's original `needs:` contained a failed FG dep, the BG job should be marked `DepFailed` rather than run. Thread the FG outcomes into `CoordinatorState`; pre-mark the relevant BG jobs as `DepFailed` before the wave loop starts so the existing cascade fires on transitive dependents.

Regression test: `bg_missing_dep_in_needs_returns_error` should be replaced by a positive test that exercises the partitioner + coordinator round-trip end-to-end (BG job with `needs:` pointing into the FG slice runs successfully after FG completes; fails-to-`DepFailed` if the FG dep fails).

Independently worth fixing in a follow-up: the unconditional "N background job(s) running" message should either wait for the coordinator to write its first `JobRow` or surface a startup-error breadcrumb. Today a coordinator that exits 1 for any reason leaves zero user-visible trace.

avihut:
## Summary

- BG hook job with `needs:` on a FG job silently never ran — partitioner left the FG name in the BG `JobSpec.needs`, coordinator's `DagGraph::new` rejected the dangling reference as `MissingDependency`, and the coordinator child exited 1 with stderr `/dev/null`'d while the parent unconditionally printed `⟳ N background job(s) running`.
- This is the exact pattern documented in `SKILL.md:506–516` and used in this repo's own `daft.yml` (`mise dev` is bg, `needs: [bun install]` which is fg).

## Fix

Three small layered changes:

1. **`partition.rs`** — strip FG names from BG `needs:`. FG-before-BG sequencing already satisfies the "happens-after" intent; the names are noise in the BG-only DAG.
2. **`coordinator/process.rs`** — `CoordinatorState::prefailed_jobs` is the new channel for "this BG job's FG dep failed; record it as skipped without running it." `run_all_with_cancel` marks these as `DepFailed` before the wave loop and reuses the existing transitive cascade.
3. **`yaml_executor/mod.rs`** — after FG results, derive the prefailed list from the ORIGINAL (pre-partition) specs and thread it into `CoordinatorState`. Inline debug paths (`DAFT_NO_BACKGROUND_JOBS`, non-Unix fallback) get the same cascade via a small `run_bg_inline_with_prefailed` helper that synthesizes `Skipped` results without spawning.

Wire-format compatibility: `prefailed_jobs` is `#[serde(default)]` so older parents that don't populate the field round-trip through the `__coordinator <state-file>` JSON payload as an empty list.

## Test plan

- [x] `partition::test_partition_bg_needs_fg_is_stripped` — partitioner-side regression
- [x] `partition::test_partition_bg_needs_mix_keeps_bg_drops_fg` — mixed BG+FG deps cleaned correctly
- [x] `coordinator::process::tests::prefailed_job_recorded_as_skipped_without_running` — prefailed name lands as `status="skipped"` in SQLite, no child process spawned
- [x] `coordinator::process::tests::prefailed_cascade_skips_transitive_bg_dependents` — transitive BG dependents cascade
- [x] `coordinator::process::tests::prefailed_does_not_block_unrelated_bg_jobs` — sibling BG jobs unaffected
- [x] `hooks::yaml_executor::tests::test_bg_needs_failed_fg_skipped_in_inline_path` — end-to-end via the inline debug path; uses a filesystem marker because the executor routes job stdout through the log sink
- [x] `mise run fmt` clean
- [x] `mise run clippy` clean (0 warnings)
- [x] `mise run test:unit` — 1822 passed

## Out of scope / follow-up

The unconditional `⟳ N background job(s) running` message in `yaml_executor/mod.rs:450` will still lie about runtime if the coordinator fails to start for some other reason (OS error, schema mismatch, etc.). That deserves its own follow-up — either wait for the coordinator to write its first `JobRow` before printing, or stop `/dev/null`'ing the coordinator's stderr.

Fixes #556.

### DIFF
--- a/src/coordinator/process.rs
+++ b/src/coordinator/process.rs
@@ -73,6 +73,19 @@ pub struct CoordinatorState {
     pub trigger_command: String,
     pub hook_type: String,
     pub worktree: String,
+    /// Names of background jobs that must be recorded as `DepFailed`
+    /// before the wave loop starts. Populated by the caller from the
+    /// foreground `JobResult`s: a BG job whose original (pre-partition)
+    /// `needs:` referenced a foreground job that did not succeed lands
+    /// here so it is not run, and its dependents cascade through the
+    /// same `DepFailed` path that handles runtime failures.
+    ///
+    /// `#[serde(default)]` is load-bearing: older parents that haven't
+    /// learned this field serialize a payload without it, and a freshly
+    /// spawned coordinator deserializes them as an empty list (i.e. no
+    /// prefailed jobs).
+    #[serde(default)]
+    pub prefailed_jobs: Vec<String>,
 }
 
 /// Wire format for state passed from parent to spawned `__coordinator`
@@ -94,6 +107,7 @@ impl CoordinatorState {
             trigger_command: String::new(),
             hook_type: String::new(),
             worktree: String::new(),
+            prefailed_jobs: Vec::new(),
         }
     }
 
@@ -101,6 +115,13 @@ impl CoordinatorState {
         self.trigger_command = trigger_command.to_string();
         self.hook_type = hook_type.to_string();
         self.worktree = worktree.to_string();
+        self
+    }
+
+    /// Set the list of background-job names that should be recorded as
+    /// `DepFailed` before the wave loop starts (see [`Self::prefailed_jobs`]).
+    pub fn with_prefailed(mut self, names: Vec<String>) -> Self {
+        self.prefailed_jobs = names;
         self
     }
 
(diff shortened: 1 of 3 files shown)


avihut:
Addressed in 7cad65de:

- **Inline cascade bug** (medium): new `expand_prefailed_closure` helper computes the transitive closure of `prefailed_bg` through the BG→BG `needs:` graph before partitioning into skipped/run sets. Without this, the `fg-fail → bg-dep → bg-transitive` chain in the inline path errored out with `MissingDependency` instead of cascading cleanly. The coordinator path was already correct via its own DFS in `run_all_with_cancel`.
- **Inflated "N background job(s) running" count** (minor): subtract `prefailed_jobs.len()` from `bg_count`; skip the message entirely when the post-subtraction count is zero.
- **Test env-var parallelism risk** (minor): replaced bare `unsafe { env::set_var }` with a `serial_test::serial(daft_no_background_jobs)`-gated RAII guard (`NoBackgroundJobsEnv`) that restores the var on drop, including on panic. The new transitive-cascade test reuses the same serial key.

New regression test `test_bg_needs_chain_cascade_skips_transitive_in_inline_path` covers the three-job chain end-to-end via filesystem markers.

Heads-up on the one CI fail on this push: `integration-tests (gitoxide, yaml)` failed on `hooks/bg-output-silent.yml` step 4 ("silent + success → log file deleted"). The scenario doesn't exercise any code path my changes touched — no `needs:`, no FG failures, no cross-partition deps — and it passed locally on first try (`mise run test:manual -- --ci hooks:bg-output-silent` → 1/1 passed). Reran the failed job and it passed clean. Looks like a timing flake on that scenario's "silent + success log removal" assertion; worth flagging separately if it recurs but not blocking this PR.

Issue #561, opened 2026-05-26
perf(test-runner): amortize daft binary startup cost across steps


avihut:
## Goal

Amortize the cost of starting the daft binary across the steps of a scenario, so the suite stops paying ~33ms × 2,217 = ~72s of serial fork+exec+initialization overhead. Largest remaining unmapped cost in #509 per the per-phase bench data in #511's PR (~18% of total serial CPU).

Part of #509.

## Current state

Every step does a fresh `daft …` invocation. The cold-cache cost is bounded by what daft does at startup, which today is non-trivial:

- clap parse + dispatch
- `dirs` resolution of XDG paths
- Store open (SQLite WAL setup + PRAGMAs in `src/store/connection.rs::bring_up`)
- Config load (`core::settings`, `core::config`)
- Update-check / trust-prune background spawn checks (`update_check::maybe_check`, `trust_prune::maybe_prune_trust`) — these themselves are guarded by `skip_startup_tasks_for`, but only for `shell-init` / `completions` / `__*` paths; the test runner's daft invocations don't qualify, so the cache-stat hits land per step.

Floor measurement (release build, M1 Max, warm FS cache):

- `time ./target/release/daft --help` → **33ms**

Across 2,217 steps that's **~72s of serial cost** that has nothing to do with what daft is actually being asked to do — it's pure per-step startup tax.

## Approach

Three architecturally different options, in increasing order of structural change:

**(a) Suppress daft's startup-time background tasks under a runner-only env var.** Cheapest. Add a `DAFT_MANUAL_TEST=1` flag that the runner already sets (or extend the existing test-mode flags) to also short-circuit `update_check::maybe_check`, `trust_prune::maybe_prune_trust`, and any other startup-time bookkeeping. Doesn't change the multicall binary model — just makes startup leaner under tests. Expected gain: ~5–10ms per invocation. ~15-25s wall-clock at jobs=1, ~3-5s at jobs=10. Lowest risk, fastest to ship.

[...]

Issue #560, opened 2026-05-26
perf(test-runner): exec daft directly when no shell features are needed


avihut:
## Goal

Exec daft (and `git-worktree-*` symlinks) directly when a step's command is a plain invocation with no shell features. Drop the `bash -c` wrapper cost from the step phase — ~38s of serial work across the suite, ~9% of total serial CPU per the instrumented bench in [#511's analysis].

Part of #509.

## Current state

`xtask/src/manual_test/daft_executor.rs:137-143` unconditionally wraps every step in `bash -c "<command>"`:

```rust
let output = Command::new("bash")
    .process_group(0)
    .args(["-c", &expanded])
    .current_dir(cwd)
    .envs(self.build_env(sandbox))
    .output()
    .with_context(|| format!("Failed to execute: {expanded}"))?;
```

Floor measurements (release build, M1 Max, warm cache):

- `time ./target/release/daft --help` → **33ms**
- `time bash -c "./target/release/daft --help"` → **50ms**

So `bash` itself contributes ~17ms × 2,217 steps = **~38s serial / 9% of suite CPU**. At `--jobs 10` with ~77% contention inflation that's ~7s of wall-clock saved on a 75s run.

## Approach

Parse the expanded command with `shlex::split` (already a daft dep — `Cargo.toml:129`). If splitting succeeds and the resulting argv is a plain invocation with no shell features, exec the program directly. Otherwise fall back to `bash -c`.

A command qualifies for the direct-exec fast path if:

- `shlex::split` returns `Some(argv)` with at least one element.
- The original command string contains none of: `|`, `>`, `<`, `&`, `;`, `` ` ``, `$(`, `*`, `?`, `[`.
  (Cheap byte-scan before shlex — quotes are fine as long as nothing inside them is a metachar.)
- The first argv element resolves to an executable on `PATH` (or is an absolute path).

Both code paths use the same `process_group(0)` + env + cwd setup. The `expand_vars` step on the original command string is unchanged (variables are resolved before the shell-features check).

Sketch:

```rust
fn try_direct_argv(command: &str) -> Option<Vec<String>> {
    if command.contains(['|', '>', '<', '&', ';', '`', '*', '?', '['])
        || command.contains("$(")
    {
        return None;
    }
    shlex::split(command).filter(|argv| !argv.is_empty())
}

impl CommandExecutor for DaftCommandExecutor {
[...]

Issue #559, opened 2026-05-26
perf(test-runner): interleave scenarios across workers to flatten the parallel tail


avihut:
## Goal

Interleave or shuffle the scenario list before parallel dispatch so the wall-clock tail of `mise run test:manual` isn't dominated by whichever worker happened to draw the slowest contiguous namespace chunk.

Part of #509.

## Current state

`xtask/src/manual_test/mod.rs:1270` sorts `scenarios` alphabetically before handing them to `par_iter().for_each_with(...)` (mod.rs:570-577). Rayon's `par_iter` partitions the slice into contiguous chunks per worker — Worker 0 gets `[0, chunk)`, Worker 1 gets `[chunk, 2·chunk)`, etc. — and only steals when a worker finishes its chunk.

Combined with the alphabetical sort, this concentrates same-namespace scenarios on the same worker. Today's namespace counts (`xtask manual-test --list | awk -F: '{print $1}' | sort -u | …` collapsed):

| Namespace | Count |
| --- | --- |
| layout | 75 |
| hooks | 69 |
| merge | 65 |
| list | 53 |
| clone | 38 |
| checkout | 33 |
| branch-delete | 30 |
| checkout-branch | 23 |
| prune | 22 |
| fetch | 22 |

With 10 workers and 581 scenarios sorted alphabetically, the initial dispatch puts e.g. all 33 `checkout:*` + all 23 `checkout-branch:*` on one or two workers in a contiguous run. If those scenarios are heavier on average than (say) `list:*` or `prune:*`, the tail of the suite is one worker grinding through their backlog while the others are idle. Field-observed pattern: the last 20–30s of a 1m13s run on M1 Max shows a single worker emitting `Checkout …` outputs serially while every other worker has long since exited.

## Approach

Three options, ordered by ROI × code-touch:

**(a) Round-robin interleave by namespace.** Group scenarios by namespace, then weave: `branch-delete:1, checkout:1, hooks:1, layout:1, …, branch-delete:2, checkout:2, …`. Distributes any per-namespace cost variance across all workers' initial chunks. ~15 lines in `mod.rs` between the discovery sort and the `par_iter`.

**(b) Random shuffle (seeded).** Cheaper to implement than (a) — just `scenarios.shuffle(&mut rng)` with a fixed seed for reproducibility. Doesn't guarantee even namespace distribution but eliminates the alphabetical clustering.

[...]

Issue #513, opened 2026-05-16
perf(test-runner): share fixture remotes across scenarios


avihut:
## Goal

Generate each unique fixture repo once at the start of the manual-test run, then clone (clonefile-ed) it into each scenario's sandbox. Today `repo_gen::generate_repo` runs once per scenario per repo — the same fixture is rebuilt N times.

Part of #509. Conceptually depends on the clonefile/reflink sub-task for the per-scenario clone-in to be O(1).

## Current state

- `xtask/src/manual_test/mod.rs:124-128` calls `repo_gen::generate_repo(repo_spec, &test_env.remotes_dir)` for every repo in every scenario.
- Many scenarios reference the same fixtures via `use_fixture:` (see `tests/manual/fixtures/repos/`).
- Each fixture is independent of the others — no cross-scenario state, so cached generation is safe.

## Approach

1. Before the scenario loop, walk all scenarios and collect the set of unique `(fixture_name, name)` tuples referenced anywhere in the suite.
2. Build a "fixture cache" directory at a known location (under the sandbox base, so it lives and dies with the run by default).
3. For each unique tuple, generate the repo once into the cache.
4. Per-scenario `register_remote` then clones (clonefile-ed) the cached repo into the scenario's `remotes/` directory.
5. Inline (non-fixture) repo specs continue to be generated per-scenario — they're scenario-specific by construction.

## Open questions

- **Cache lifetime: per-run or persistent across runs?** Per-run is the safer default; persistent is the speedier upgrade. Ship per-run first and measure the gain. If the cache build is still a bottleneck, add a persistent mode behind a flag with a hash-based invalidation key (hash the fixture YAML + the substituted `{{NAME}}`).
- **Where does the cache live?** Inside `base_dir/fixture-cache/` so cleanup is automatic. With the RAM-disk sub-task, the cache is on RAM too — clones are double-fast.

## Measurement

Use `mise run bench:test:manual` (filed under #509 as the benchmark-harness prerequisite). The PR description must include:

- Count of `repo_gen::generate_repo` invocations before vs after on the full suite. With shared fixtures, this should drop to one per unique fixture rather than one per (fixture × scenario).
- Full-suite wall-clock before/after.
- Per-scenario p50 — heavy-fixture scenarios should shift significantly faster while inline-repo scenarios stay flat.

## Acceptance criteria

- [ ] Each unique fixture is generated exactly once per run, regardless of how many scenarios reference it.
[...]

PR #562, opened 2026-05-26
perf(test-runner): use APFS clonefile / Linux FICLONE for sandbox template snapshots

### LINKED ISSUE #511: perf(test-runner): use APFS clonefile / reflink for template snapshots
## Goal

Replace `cp -a` with clonefile / reflink for the per-scenario template snapshot and reset in `xtask/src/manual_test/env.rs:184-225`. On APFS this turns the snapshot into an O(1) copy-on-write operation; on supporting Linux filesystems (`btrfs`, `xfs` with reflink, `zfs`) `cp --reflink=auto` does the same.

Part of #509. Independent of the parallel-execution sub-task — can ship first as a warm-up.

## Current state

`env.rs:185` runs `cp -a remotes/ remotes-template/` for every scenario, which is a full byte copy of every fixture's git objects. Same pattern in `reset()` (`env.rs:215`) when restoring from the template.

## Approach

Platform-detect at runtime and pick the right `cp` invocation:

- macOS: `cp -c -R src dst` — the `-c` flag enables clonefile, available since macOS 10.13.
- Linux: `cp --reflink=auto -a src dst` — if the underlying FS does not support reflinks, `auto` silently falls back to a normal copy, no breakage.
- Other / unknown: fall back to `cp -a` (today's behavior).

Implement as a small `clone_dir(src, dst) -> Result<()>` helper in `env.rs`. Probe the platform once at module load (or per-call — the syscall cost is negligible).

## Measurement

Use `mise run bench:test:manual` (filed under #509 as the benchmark-harness prerequisite). The PR description must include:

- Baseline (cp -a) wall-clock for the template-snapshot step alone (per-scenario × N scenarios).
- After-clonefile wall-clock for the same.
- Full-suite wall-clock before/after — clonefile should shave a measurable chunk even before parallel execution lands.
- Confirmation on Linux that `--reflink=auto` falls back transparently on non-reflink filesystems (run the bench on both ext4 and a reflink-capable FS if possible).

## Acceptance criteria

- [ ] Template snapshot uses clonefile on macOS APFS / reflink on supporting Linux filesystems.
- [ ] Fallback to `cp -a` on unsupported filesystems is silent and correct.
- [ ] Measurable wall-clock improvement on the template-snapshot step for fixtures with non-trivial git history (reported via `bench:test:manual`).
- [ ] Measurable wall-clock improvement on the full suite even with the existing serial scheduler.
- [ ] No regression on platforms without clonefile / reflink support.
- [ ] Unit test or scenario covers the clone-then-mutate-source pattern to confirm CoW semantics work as expected.

avihut:
## Summary

- New `daft::cow_copy::{copy_file, copy_dir}` primitive — reflink-or-byte-copy via the `reflink-copy` crate. Same primitive #387's `copy_paths:` will consume in production daft code.
- xtask's manual-test sandbox uses an inlined sibling helper (`xtask::manual_test::cow_copy`) instead of `cp -a`, keeping `runner.rs / sandbox.rs / executor.rs` free of `daft::*` imports per the YAML runner port rule in CLAUDE.md.
- Per-phase timing instrumentation added under `DAFT_MANUAL_TEST_EMIT_TIMING` (`setup_ms`, `fixture_ms`, `template_ms` alongside `elapsed_ms`) so future #509 perf work can attribute wall-clock to the right phase.

Fixes #511.

## What changed

- `Cargo.toml` — adds `reflink-copy = "0.1.29"`. Crate published 2025-03-04, well past the 7-day cooldown. Public API is fully safe (`reflink`, `reflink_or_copy`, `check_reflink_support` — no `unsafe fn`), preserving daft's `forbid(unsafe_code)` posture. Same precedent as `rusqlite` over `heed`.
- `src/cow_copy.rs` (new) + `src/lib.rs` — `pub mod cow_copy;`. File-level + dir-level primitives with reflink-or-byte-copy fallback, mode-bit preservation, symlink recreation. 7 unit tests cover CoW independence, mode preservation, symlink handling, and error contract.
- `xtask/Cargo.toml` — adds `reflink-copy` and `walkdir` as xtask's own deps.
- `xtask/src/manual_test/cow_copy.rs` (new) — inlined sibling using the same primitives. The ~30 lines of duplication is intentional and documented in the module's docstring (CLAUDE.md's YAML runner port rule forbids runner-core from naming `daft::*` so the spin-out invariant holds).
- `xtask/src/manual_test/sandbox.rs` — `Sandbox::create_template` and `Sandbox::reset` switch from `Command::new(\"cp\").args([\"-a\"])` to `cow_copy::copy_dir`. The `.process_group(0)` SIGINT-insulation plumbing is gone (no subprocess to insulate). Adds an end-to-end test exercising the `create_template` → mutate → `reset` lifecycle.
- `xtask/src/manual_test/mod.rs` — `run_one_scenario_inner` now times each phase and emits `setup_ms` / `fixture_ms` / `template_ms` on the `[bench]` line alongside `elapsed_ms` when `DAFT_MANUAL_TEST_EMIT_TIMING=1`. The percentiles script in `benches/scenarios/test_manual_scale.sh` is backwards-compatible (it greps `elapsed_ms=` then takes `\$1`).

## Issue #387 amended

#387 (`copy_paths:` in `daft.yml`) is now wired to consume `daft::cow_copy::{copy_file, copy_dir}` directly — its body has been updated to reframe its scope as adding policy (`prefer | only | never`), `max_size` filtering, and the daft.yml surface on top of #511's primitives.

## Measurement

Full manual-test suite: **581/581 passing** on macOS APFS / M1 Max.

Wall-clock numbers (`mise run bench:tests:manual:scale BENCH_JOBS=10 BENCH_RUNS=3`):

| | This PR | Pre-PR (post-#510 baseline at jobs=10) |
| --- | ---:| ---:|
| Mean | 88.5s ± 6.9s | 82.2s (single trial) |
| Min | 84.4s | — |

**Net wall-clock effect at jobs=10 is within hyperfine's noise floor.** Per-phase instrumentation explains why: template snapshot is only 2.6s / 0.6% of the suite's serial CPU work, so clonefile-ing it shaves milliseconds that get lost in parallel scheduling noise.

### Per-phase breakdown (serial, with this PR's instrumentation, 579 fixture-using scenarios sampled at jobs=1)

| Phase | Cumulative | % | p50 | p95 | max |
| --- | ---:| ---:| ---:| ---:| ---:|
| Step (`elapsed`) | 286.7s | 70.3% | 476ms | 1037ms | 3170ms |
| Fixture gen | 118.8s | 29.1% | 241ms | 265ms | 959ms |
| Template snapshot | 2.6s | 0.6% | 5ms | 6ms | 23ms |
| Setup (`create_at`) | 0.0s | 0.0% | 1ms | — | 11ms |

The clonefile path is engaging and dropping template snapshot to a 5ms median. The dominant remaining costs are fixture generation (which #513 targets — its hot path becomes O(1) with the helper this PR introduces) and the step phase (subject of #560/#561).

## Follow-up issues opened

The bench data surfaced three actionable bottlenecks not previously mapped under #509. All three are filed, milestoned for Public Launch, and parked in the daft project's **Ready** column:

- **#559** — `perf(test-runner): interleave scenarios across workers to flatten the parallel tail`. Diagnosed during PR review: rayon's `par_iter` chunks the alphabetically-sorted scenario list contiguously, clustering same-namespace scenarios on the same worker.
- **#560** — `perf(test-runner): exec daft directly when no shell features are needed`. Floor measurement showed `bash -c \"daft …\"` adds ~17ms × 2,217 steps = ~38s serial overhead.
- **#561** — `perf(test-runner): amortize daft binary startup cost across steps`. Floor measurement: ~33ms × 2,217 = ~72s serial of pure daft startup tax.

The expected stacking — #511 + #512 + #513 + #559 + #560 + #561(a) — should land the suite at ~35-45s wall-clock at jobs=10, down from today's ~75-88s. #561(b) (long-lived daft daemon per sandbox) is gated on whether the cheaper wins suffice.

## Test plan

- [x] Unit tests for `daft::cow_copy::{copy_file, copy_dir}` — CoW independence, mode-bit preservation, symlink handling, error contract.
- [x] Unit tests for `xtask::manual_test::cow_copy::copy_dir` — same coverage on the inlined sibling.
- [x] `Sandbox::create_template` → mutate → `reset` lifecycle test in `sandbox.rs::tests` (the previously-uncovered `reset()` path).
- [x] CLAUDE.md runner-core rule: `rg 'daft::|DAFT_' xtask/src/manual_test/{runner,sandbox,executor}.rs` returns zero matches outside docstrings.
- [x] Full manual-test suite: 581/581 passing on macOS APFS.
- [x] Clippy zero-warnings, fmt clean (lefthook pre-commit enforced).
- [ ] CI / Linux validation — exercised via `cp --reflink=auto`-equivalent fall-through (Linux runners are typically on ext4 which doesn't reflink; the silent byte-copy fallback path is what gets exercised).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -741,6 +741,7 @@ dependencies = [
  "r2d2",
  "r2d2_sqlite",
  "ratatui",
+ "reflink-copy",
  "regex",
  "rusqlite",
  "rusqlite_migration",
@@ -3733,6 +3734,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "reflink-copy"
+version = "0.1.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13362233b147e57674c37b802d216b7c5e3dcccbed8967c84f0d8d223868ae27"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "rustix",
+ "windows",
+]
+
+[[package]]
 name = "regex"
 version = "1.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5196,6 +5209,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 
 [[package]]
+name = "windows"
+version = "0.62.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "527fadee13e0c05939a6a05d5bd6eec6cd2e3dbd648b9f8e447c6518133d8580"
+dependencies = [
+ "windows-collections",
+ "windows-core",
+ "windows-future",
+ "windows-numerics",
+]
+
+[[package]]
+name = "windows-collections"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b2d95af1a8a14a3c7367e1ed4fc9c20e0a26e79551b1454d72583c97cc6610"
+dependencies = [
+ "windows-core",
+]
+
+[[package]]
 name = "windows-core"
 version = "0.62.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5206,6 +5240,17 @@ dependencies = [
  "windows-link",
  "windows-result",
  "windows-strings",
+]
+
+[[package]]
+name = "windows-future"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1d6f90251fe18a279739e78025bd6ddc52a7e22f921070ccdc67dde84c605cb"
+dependencies = [
+ "windows-core",
+ "windows-link",
+ "windows-threading",
 ]
 
 [[package]]
@@ -5235,6 +5280,16 @@ name = "windows-link"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f0805222e57f7521d6a62e36fa9163bc891acd422f971defe97d64e70d0a4fe5"
+
+[[package]]
+name = "windows-numerics"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e2e40844ac143cdb44aead537bbf727de9b044e107a0f1220392177d15b0f26"
+dependencies = [
+ "windows-core",
+ "windows-link",
+]
 
 [[package]]
 name = "windows-result"
@@ -5360,6 +5415,15 @@ dependencies = [
  "windows_x86_64_gnu 0.53.1",
  "windows_x86_64_gnullvm 0.53.1",
  "windows_x86_64_msvc 0.53.1",
+]
+
+[[package]]
+name = "windows-threading"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3949bd5b99cafdf1c7ca86b43ca564028dfe27d66958f2470940f73d86d75b37"
+dependencies = [
+ "windows-link",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5751,10 +5751,12 @@ dependencies = [
  "indicatif",
  "nix",
  "rayon",
+ "reflink-copy",
  "serde",
  "serde_yaml",
  "tempfile",
  "term-styles",
+ "walkdir",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,6 +130,14 @@ shlex = "1.3"
 uuid = { version = "1", features = ["v7"] }
 fs2 = "0.4"
 walkdir = "2.5"
+# Copy-on-write file copy (APFS clonefile on macOS, ioctl(FICLONE) on Linux
+# btrfs/XFS/ZFS, ReFS block-clone on Windows). Falls back to byte-copy on
+# unsupported filesystems via `reflink_or_copy`. Used by `src/cow_copy.rs`
+# and consumed from xtask's manual-test runner; future home of #387's
+# `copy_paths:` implementation. Public API is fully safe (`unsafe` is
+# confined to the crate's internal sys layer), so `forbid(unsafe_code)`
+# stays intact — same precedent as `rusqlite` over `heed`.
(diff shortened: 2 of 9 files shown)


avihut:
Thanks for the thorough review. Triage:

## Critical: xtask Windows compile — **declined** with evidence

Verified against the actual `windows-check` log for the last passing run on this PR (job [77870523952](https://github.com/avihut/daft/actions/runs/26450419534/job/77870523952)). The `cargo check --all-targets` step compiles **only** the `daft` package, not `xtask` or `term-styles`:

```
Compiling daft v1.14.4 (D:\a\daft\daft)   ← only workspace member compiled
```

83 compile lines total in that job, none for xtask or term-styles. This is because the workflow runs `cargo check` from the workspace root where `[package] name = "daft"` lives — cargo scopes to the root package by default unless `--workspace` is explicit or `default-members` is set.

This means the unguarded Unix imports in `xtask/src/manual_test/cow_copy.rs` and the `PermissionsExt` use in `sandbox.rs::test_create_template_and_reset_lifecycle` are **not** broken by current CI, and the existing master code in `xtask/src/manual_test/daft_executor.rs:20` (`use std::os::unix::process::CommandExt;` — also unconditional) has been passing Windows checks for the same reason.

Gating just the new xtask cow_copy items while leaving `daft_executor.rs`, `repo_gen.rs`, etc. unguarded would create inconsistency. The xtask test runner is Unix-only by design (uses `process_group`, `/tmp` paths, the manual-test sandbox layout) — making just one module of it Windows-portable isn't useful when the rest of the crate can't compile on Windows anyway.

If we want a real Windows story for the xtask runner (which the YAML-runner spin-out memory in CLAUDE.md doesn't seem to call for), that's its own scoped project, not piecemeal cfg-gates in this PR. Until then, the daft-side `cow_copy.rs` is what carries the Windows promise (for #387's `copy_paths:`), and it's already correctly gated by commit `11e80cd`.

## Minor: test coverage gap — **addressed** in `f52932cf`

Added `errors_when_source_is_a_file` to `xtask/src/manual_test/cow_copy.rs::tests`, mirroring the daft side. Test count now 4 (was 3); 4/4 pass locally.

## Observation: TOCTOU in `copy_file` — **acknowledged, no action**

The window between `reflink_or_copy(src, dst)` and `symlink_metadata(src)` is real but not exploitable in the current callsites (sandbox-controlled paths). If a future caller hits a less-trusted source path, the right fix is reading metadata once before the copy and using it for both — not gating now without a real consumer pushing the requirement. Noted for #387 review.

Issue #509, opened 2026-05-16
perf(test-runner): speed up the YAML scenario suite


avihut:
## Goal

`mise run test:manual --ci` (the full YAML scenario suite) takes a very long time locally even on heavily-specced machines. Running tests from two worktrees simultaneously buckles the machine. This makes day-to-day daft development slow and discourages running the YAML suite locally as part of normal flow.

This umbrella tracks a focused performance overhaul of `xtask/src/manual_test/` (the runner) and its surrounding infrastructure (sandbox layout, fixture handling, cross-worktree binary cost).

## Current state

- `xtask/src/manual_test/mod.rs:102` runs scenarios in a strictly serial `for` loop — zero parallelism, 573 scenarios.
- Each scenario gets its own sandbox under `/tmp/daft-manual-test-<ts>/` (`env.rs:64`). Sandboxes are already isolation-clean, so parallel execution is correctness-safe at the runner layer.
- Template snapshot uses `cp -a` (`env.rs:185`) — a full byte copy of every fixture's git objects per scenario.
- `repo_gen::generate_repo` runs per scenario — the same fixture is rebuilt N times across the suite.
- Running tests in two worktrees doubles the cost of building `target/release/daft` and competes for the same `/tmp` IO + macOS file cache.

## Sub-tasks (ordered by ROI × independence)

- [ ] #510 — Parallelize scenario execution (rayon, configurable concurrency cap)
- [ ] #559 — Interleave scenarios across workers to flatten the parallel tail (post-#510; namespace-aware round-robin instead of alphabetical contiguous chunks)
- [ ] #516 — Introduce `CommandExecutor` port at the runner/daft seam (groundwork for the perf changes below)
- [ ] #518 — Legible test-runner output (scenario PASS/FAIL footers, failed-scenarios summary, verbosity ladder)
- [ ] #511 — APFS clonefile (`cp -c`) for template snapshot + reset, with `--reflink=auto` fallback on Linux
- [ ] #560 — Exec daft directly when no shell features are needed (drop ~17ms `bash -c` wrapper × ~2,217 steps; depends on #516's port surface)
- [ ] #561 — Amortize daft binary startup cost across steps (~33ms × ~2,217 steps; lands after #560 so each measurement is uncontaminated)
- [ ] #512 — RAM-disk sandbox via `DAFT_MANUAL_TEST_BASE` + `mise run test:manual:ramdisk` helper
[...]

PR #564, opened 2026-05-27
perf(test-runner): interleave scenarios across workers to flatten the parallel tail

### LINKED ISSUE #559: perf(test-runner): interleave scenarios across workers to flatten the parallel tail

avihut:
## Summary

- Round-robin interleave by namespace before dispatching to rayon's `par_iter`, breaking the alphabetical clustering that concentrated heavy namespaces (layout, hooks, merge) on single workers.
- Reporter output ordering preserved: `scenario_files` stays alphabetical; a new `dispatch_order: Vec<usize>` carries the interleaved permutation; `ScenarioOutcome.index` keeps its alphabetical-position semantics so the post-run sort still produces alphabetical non-TTY summary.
- `DAFT_MANUAL_TEST_DEBUG_DISPATCH=1` (pair with `CI=1` or `NO_PROGRESS=1`) prints `[dispatch] thread=… idx=… ns=…` per scenario for verifying the per-worker mix.

Closes #559. Part of #509 umbrella.

## Bench

Alternating hyperfine, 1 warmup + 4 runs each, M1 Max, `--jobs 10`:

| | mean [s] | σ [s] | min [s] | max [s] |
|---|---:|---:|---:|---:|
| master `16f6b9ff` | 73.328 | ±3.435 | 70.954 | 78.393 |
| this branch | **70.643** | **±0.310** | 70.257 | 70.926 |
| Δ | **-2.69 (-3.7%)** | **-91%** | | |

The headline is the **variance reduction** (σ 3.4s → 0.3s, 11x tighter) — the mechanism #559 predicted. Mean drops modestly (3.7%) but the operationally meaningful result is that master's *best* run (70.95s) is essentially tied with this branch's *worst* (70.93s) — the change pins the suite to its best-case rather than letting it wander up to ~78s on a bad trial.

Earlier non-alternating benches inflated the mean win because of thermal-order confounding on M1 Max — the second-run side ran hot. Alternating trials eliminates that.

The CI runners are different hardware (shared / burst CPUs) where worker imbalance hits harder, so the variance signal here may translate to either a bigger mean drop OR a tighter run-to-run distribution in the workflow numbers. Worth a glance once the checks land.

## Acceptance criteria (#559)

- [x] Per-worker scenario assignment interleaved across namespaces (verified via `DAFT_MANUAL_TEST_DEBUG_DISPATCH` — 40 dispatch lines spanned 16+ distinct namespaces across all 10 threads).
- [x] Wall-clock at `--jobs 10` improves over master baseline (alternating bench).
- [x] Determinism preserved (BTreeMap + VecDeque round-robin, no RNG; covered by `interleave_is_deterministic` unit test).
- [x] No regression in reporter ordering (empirically verified on a 581-scenario run; first ~50 lines remain alphabetical by `namespace:stem`).

## Test plan

- [x] `mise run fmt:check`, `mise run clippy`, `mise run test:unit` — clean.
- [x] `cargo test -p xtask manual_test::` — clean (131 passed, includes 10 new dispatch-order unit tests).
- [x] `mise run test:manual` — 581/581 scenarios pass.
- [x] Solo branch run timed at 69s (cold); alternating bench averages 70.6s under sustained load.
- [ ] CI workflow checks land green and (bonus) we get a second hardware data point on the variance behavior.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,7 +199,7 @@ jobs:
     if: >-
       !contains(github.event.pull_request.labels.*.name, 'release') &&
       (needs.changes.outputs.self == 'true' || needs.changes.outputs.rust ==
-      'true')
+      'true' || needs.changes.outputs.xtask == 'true')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -225,7 +225,7 @@ jobs:
     if: >-
       !contains(github.event.pull_request.labels.*.name, 'release') &&
       (needs.changes.outputs.self == 'true' || needs.changes.outputs.rust ==
-      'true')
+      'true' || needs.changes.outputs.xtask == 'true')
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v6
@@ -337,7 +337,8 @@ jobs:
     if: >-
       !contains(github.event.pull_request.labels.*.name, 'release') &&
       (needs.changes.outputs.self == 'true' || needs.changes.outputs.rust ==
-      'true' || needs.changes.outputs.tests == 'true')
+      'true' || needs.changes.outputs.tests == 'true' ||
+      needs.changes.outputs.xtask == 'true')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -282,9 +282,10 @@ Hard rules:
   evaluation are candidates for future ports. Extracting them speculatively is
   the trap; wait for a downstream PR with a concrete reason.
 - **Intentional exception:** `DAFT_MANUAL_TEST_BASE`,
-  `DAFT_MANUAL_TEST_EMIT_TIMING`, and `DAFT_MANUAL_TEST_JOBS` are the runner's
-  own config knobs and live in `mod.rs` / `main.rs`. Renaming the namespace is a
-  spin-out concern, not a coupling concern.
+  `DAFT_MANUAL_TEST_EMIT_TIMING`, `DAFT_MANUAL_TEST_JOBS`, and
+  `DAFT_MANUAL_TEST_DEBUG_DISPATCH` are the runner's own config knobs and live
+  in `mod.rs` / `main.rs`. Renaming the namespace is a spin-out concern, not a
+  coupling concern.
 
 **Design language**: `xtask/src/manual_test/reporter/CLAUDE.md` constrains the
 output's appearance (color budget, hierarchy, iconography, microcopy, rhythm).
--- a/xtask/src/manual_test/mod.rs
+++ b/xtask/src/manual_test/mod.rs
@@ -581,7 +581,20 @@ fn run_parallel(
     // can verify per-thread namespace mix without altering normal output.
     // Read once outside the closure — `std::env::var_os` synchronizes on a
     // global lock, no point paying for it per scenario.
-    let debug_dispatch = std::env::var_os("DAFT_MANUAL_TEST_DEBUG_DISPATCH").is_some();
+    //
+    // Suppress on TTY: `eprintln!` from worker threads races indicatif's
+    // live-region redraws and garbles the terminal. Pair the env var with
+    // `CI=1` or `NO_PROGRESS=1` (both already disable `show_progress`) to
+    // see the lines cleanly. If the user set the env on a TTY, warn once so
+    // they know why nothing is appearing.
+    let debug_dispatch_requested = std::env::var_os("DAFT_MANUAL_TEST_DEBUG_DISPATCH").is_some();
+    let debug_dispatch = debug_dispatch_requested && !show_progress;
+    if debug_dispatch_requested && show_progress {
+        eprintln!(
+            "[dispatch] DAFT_MANUAL_TEST_DEBUG_DISPATCH set but TTY progress is on; \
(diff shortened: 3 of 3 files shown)


avihut:
Addressed all three review nits in 29b74a43:

- `interleave_by_namespace`: held the `get_mut` reference instead of re-`get`-ing after `pop_front` — one lookup per iteration.
- Test renamed: `interleave_groups_top_level_scenarios_together` → `interleave_bare_scenarios_alternate_with_namespaced` (the original name was the opposite of what it asserted).
- `label_from_path`: added a doc comment about the one-level layout assumption (`scenarios_dir/<namespace>/<stem>.yml`); deeper paths get the immediate parent as namespace.

Behavior unchanged on all three. Tests still green.

PR #580, opened 2026-05-28
feat(test-runner): redesign the live progress region with a braille completion-map

### LINKED ISSUE #579: feat(test-runner): reorder live progress region and add progress bars
## Summary

The live progress region of the YAML manual test runner currently stacks (top → bottom):

1. Completed scenarios (streamed into scrollback, per verbosity)
2. In-flight worker rows
3. Totals / summary bar

Reorder so the totals sit **above** the in-flight worker rows:

1. Completed scenarios (scrollback, per verbosity) — unchanged
2. **Totals / summary bar**
3. **In-flight worker rows**

While reordering, restructure both lines to add real progress bars and make the
region friendlier to narrow terminals.

## Unifying rule: `[bar] → [counter] → [time]`

In **both** lines, wherever there is a progress bar it is immediately followed
by the counter of what that bar measures, and then the time counter:

1. Progress bar
2. `completed / total` counter (scenarios for totals, steps for a worker), whitespace-padded
3. Time counter (elapsed)

Everything else on the line follows after that triplet.

## Totals line (left → right)

1. Progress bar — scenario completion progression
2. `completed / total` scenarios, whitespace-padded
3. Elapsed time counter
4. `workers running / total workers allocated`, whitespace-padded
5. The rest of the data shown today (failed, cancelled, `(cancelling)`)

The totals data itself is identical to today — only its order within the line
changes, plus the leading completion bar and the new running/allocated segment.

## Worker line (one line per in-flight worker)

1. Progress bar — completed-steps progression for that scenario
2. `completed / total` steps, whitespace-padded
3. Elapsed time counter (per-scenario)
4. The scenario name
5. The current running step name — cycled and truncated for narrow displays

> Note: the requester described the worker-row field order as "right to left"
> while the totals line is "left to right". The exact left-to-right orientation
> of the worker row (whether the step bar lands on the right edge, mirroring the
> totals) is to be confirmed during planning.

## Goals

- Friendlier to narrow displays — no row may wrap (a wrapped row desyncs
  indicatif's line accounting and strands ghost rows; truncation is a
  correctness requirement, not cosmetics).
- Reuse the existing color budget (`reporter/CLAUDE.md` §1/§8) — no new color
  slots.

## Scope / non-goals

- Internal test-runner output only. No CLI surface change → no man pages, shell
  completions, or SKILL.md updates.
- `reporter/CLAUDE.md` §8 (live progress region) must be updated to match the
  new structure (it travels with the runner spin-out).
- Verbosity is orthogonal to the live region — all four tiers (`-q`/default/
  `-v`/`-vv`) render the same structure on a TTY; non-TTY suppresses it. This
  PR must preserve that.

## Touch points

- `xtask/src/manual_test/progress/indicatif_sink.rs` — reorder rows, summary
  gains a completion bar + running/allocated segment, worker rows go
  spinner → step bar, new column order + truncation.
- `xtask/src/manual_test/mod.rs` — thread total workers (`jobs`) into
  `progress_sink_for`; the pre-scan column widths and `step_label_width()` may
  change with the layout (the `render_row_msg_label_width_matches_step_label_width_formula`
  contract test is locked to that formula and changes in lockstep).

## Acceptance

- Manual TTY smoke (`mise run test:manual` with several jobs) at a normal width
  **and** at ~60 columns: no stranded/ghost rows, correct truncation, totals
  above workers.
- `mise run fmt`, `mise run clippy`, `mise run test:unit` green.

avihut:
## What

Redesigns the YAML manual-test runner's live progress region — the pinned region
shown on a TTY during a parallel run. This PR opened with a first pass (reorder
totals above the worker rows, block-style progress bars); review feedback drove a
ground-up redesign of the visual language. This description reflects the **final
state**.

**Before this PR:** completed scenarios (scrollback) → in-flight worker rows
(spinner only) → a scalar totals bar. The region churned in height as workers
ramped up and wound down, and nothing adapted to narrow terminals.

**After:** a calm, fixed-height region whose _singular_ totals line is a
distinctive cyan braille **completion-map**, with _light, quiet_ worker rows
beneath it:

```
⟨⣿⡇⣿ ⣧ … ⟩  211/581  0:05  9/10 running                    ← totals (cyan completion-map)
▬▬▬▬───────  2/5  1.2s  checkout-basic · Inspect workspace  ← busy worker
─────────────  idle                                         ← idle worker slot
```

## Totals line — a spatial completion-map (braille)

- A cyan braille **field** (`FIELD_WIDTH = 16` cells → 128 dots), framed in a dim
  `⟨ ⟩`. It lights `round(done / total × dots)` dots, so the lit fraction tracks
  completion **linearly**.
- _Which_ dots light follows a fixed **scattered** order (a gentle bottom-left →
  top-right "developing-photo" dissolve), so the field fills evenly rather than
  wiping left-to-right. A lit dot is _progress rendered spatially_ — not a
  specific finished scenario. (An earlier per-cluster scheme left the field
  near-empty until a late burst under the breadth-first scheduler; the
  count-proportional dissolve fixes that.)
- Then the stats: `done/total` scenarios → elapsed (dim) → `R/A running`
  (`A` = worker-pool size) → `M failed` (**hidden at 0**, bold-red when > 0) →
  `C cancelled` (yellow) → `(cancelling)` while interrupted. Segment separators
  are a dim `◆`.
- The totals line carries **no indicatif `{bar}`** — a count-based bar can't
  render a scattered index-map, so the field is a hand-built string set via
  `{prefix}`.

## Worker rows — fixed slot pool, calm tail

- **Fixed slot pool:** one row per worker (capped at the scenario count), sized
  once in `run_started` and **never resized**. A worker claims the lowest free
  slot on start and releases it to a quiet dim `idle` placeholder on finish. This
  holds the region height still and kills the add/remove-vs-`println` ghost-row
  race **by construction**.
- Each busy row: a medium-rect `▬`/`─` bar (default-fg fill over a dim track) →
  `done/total` steps → per-scenario elapsed (dim) → `scenario · step` flowing
  naturally (scenario name default-fg; the ` · ` separator and the changing step
  name dim, so the step doesn't flicker-grab the eye), with `(slow)` in yellow
  past 5 s. No fixed name column.
- Rows are **keyed by scenario index**, not name (scenario names are not unique).

## Persisted map + interrupt handling

- **The filled map persists past the run.** `run_finished` freezes
  `⟨…⟩  done/total scenarios` into scrollback above the reporter's summary block
  instead of letting it vanish with the cleared region. After a cancel,
  `done < total`, so the partial map shows how far the run got.
- **First Ctrl+C — soft cancel:** the region collapses cleanly, the totals line
  gains `(cancelling)`, and in-flight scenarios stream their `⊘ (cancelled)`
  footers as the run winds down (so the cancelled count and final summary print
  correctly).
- **Second Ctrl+C — hard exit:** the cooperative teardown can't run when every
  worker is wedged in a slow subprocess, so a process-global handle collapses the
  region from the signal handler; cleanup then runs and the process exits 130.
  This path **prints nothing** — no totals "end screen." (An earlier version
  printed a passed/failed breakdown here, but indicatif's count-based erase is
  off-by-one on the async `ctrlc` thread, so the live frame survives and an end
  screen below it reads as a duplicate totals line. Per maintainer call, a single
  stranded line beats a duplicate.)

## Narrow-display friendliness (a correctness property)

Each line's variable tail is a single indicatif `{wide_msg}`, which **truncates
rather than wraps** to the terminal width. A wrapped row desyncs indicatif's
line-count accounting and strands a ghost row in scrollback — the hazard the
trailer / `state_lock` / remove-then-`println` machinery exists to prevent — so
keeping every row exactly one line tall at any width is load-bearing, not
cosmetic. Truncation is ANSI-aware, so inlined color survives a cut.

## Design language & scope

- **No new color slots** (`reporter/CLAUDE.md` §1/§3/§8, rewritten). Cyan is the
  structural anchor and now marks the totals completion-field; worker bars are
  neutral (default-fg fill, dim track), step names and separators dim, yellow for
  slow/cancel, bold-red for failures.
- Internal test-runner output only — **no CLI surface change** (no man pages,
  shell completions, or SKILL.md updates).
- Non-TTY path stays **byte-identical** to before (the `NoopProgressSink`).

## Testing

- `mise run fmt`, `mise run clippy` (zero warnings), `mise run test:unit` — green.
- Unit tests cover the index→dot mapping, a dot lighting on a full cluster, the
  small-run 1-dot-per-scenario case, `render_field` for a known field, the row
  tail/counter rendering, summary-style validity, and the persisted completion
  line. Render/teardown logic is tested against a **hidden indicatif draw target**
  (no TTY, no spawned processes).
- **Manual TTY smoke** at 100 and **60 cols**: the cyan completion-map fills
  scattered, worker rows show medium-rect bars with flowing `scenario · step`
  tails, **0 rendered lines exceed the terminal width** (clean truncation, no
  wrap), and no stranded ghost rows.
- **Cancel + force-quit** verified in a real terminal: soft cancel collapses
  cleanly with `(cancelling)`; a double-Ctrl+C force-quit leaves at most a single
  summary line (never a duplicate) before the cleanup notices and exit 130.

Fixes #579

### DIFF
--- a/xtask/src/manual_test/mod.rs
+++ b/xtask/src/manual_test/mod.rs
@@ -106,31 +106,20 @@ fn max_scenario_name_width(metas: &[ScenarioMeta]) -> usize {
         .unwrap_or(0)
 }
 
-/// Compute the widest `[N/M] step_name` label across all
-/// (scenario, step) pairs. Used by the live progress region to pad the
-/// per-row step column so the elapsed counter to its right lands at a
-/// stable column across in-flight rows.
-fn max_step_label_width(metas: &[ScenarioMeta]) -> usize {
+/// Compute the widest `done/total` step counter across the discovered
+/// set, in chars. Used by the live progress region to pad each worker
+/// row's step counter so the time counter to its right lands at a stable
+/// column across in-flight rows. The widest counter for a `T`-step
+/// scenario is `T/T`, so its width is `2 * digit_count(T) + 1`.
+fn max_step_counter_width(metas: &[ScenarioMeta]) -> usize {
     metas
         .iter()
-        .flat_map(|m| {
-            let total = m.step_names.len();
-            m.step_names
-                .iter()
-                .enumerate()
-                .map(move |(i, name)| step_label_width(i + 1, total, name))
+        .map(|m| {
+            let d = digit_count(m.step_names.len());
+            2 * d + 1
         })
         .max()
         .unwrap_or(0)
-}
-
-/// Visible width of a `[N/M] step_name` label (no styling, no padding).
-/// Single helper so `max_step_label_width` and the live sink agree on
-/// the formula.
-pub(crate) fn step_label_width(idx_one_based: usize, total: usize, step_name: &str) -> usize {
-    // "[{idx}/{total}] {step_name}" — counter brackets + slash + 1 space.
-    let counter_len = digit_count(idx_one_based) + 1 + digit_count(total) + 2;
-    counter_len + 1 + step_name.chars().count()
 }
 
 fn digit_count(n: usize) -> usize {
@@ -562,10 +551,10 @@ pub fn run(
     // first scenario.
 
     // Pre-scan scenario files once for column-width sizing in the live
-    // in-flight region: pad scenario names (column 1) and the
-    // `[N/M] step_name` label (column 2) so the elapsed counter (column 3)
-    // stacks across rows. Scrollback footers don't pad — durations sit
-    // directly after the scenario name.
+    // in-flight region: pad each worker row's scenario name and its
+    // `done/total` step counter so the columns to their right (time, then
+    // the truncating step-name tail) stack across rows. Scrollback footers
+    // don't pad — durations sit directly after the scenario name.
     //
     // The scan is `peek_scenario_metadata` — a cheap text scan, not a
     // full YAML parse. ~200ms for 580 files on an SSD.
@@ -574,7 +563,7 @@ pub fn run(
         .filter_map(|p| peek_scenario_metadata(p))
         .collect();
     let name_column_width = max_scenario_name_width(&metas);
-    let step_column_width = max_step_label_width(&metas);
+    let step_counter_width = max_step_counter_width(&metas);
     let reporter = reporter::reporter_for(verbosity);
 
     // Interactive and --setup-only stay on the streaming serial path. Both
@@ -633,7 +622,7 @@ pub fn run(
     let progress = progress::progress_sink_for(
         show_progress,
         name_column_width,
-        step_column_width,
+        step_counter_width,
         jobs,
         interrupt.clone(),
     );
@@ -2271,8 +2260,7 @@ mod non_interactive_template_tests {
 #[cfg(test)]
 mod peek_scenario_metadata_tests {
     use super::{
-        max_scenario_name_width, max_step_label_width, peek_scenario_metadata, step_label_width,
-        ScenarioMeta,
+        max_scenario_name_width, max_step_counter_width, peek_scenario_metadata, ScenarioMeta,
     };
     use std::io::Write;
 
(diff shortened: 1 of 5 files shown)


avihut:
Thanks for the review — addressed in 152392e0:

- **Medium (step_started comment vs code):** restructured to read only the elapsed under the `rows` lock, drop it, format the tail/counter, then re-acquire to push the update — so concurrent workers no longer serialize on each other's string building, and the comment now matches the code.
- **Minor (NO_COLOR bypass):** documented the `\x1b[2m` raw-dim carve-out for `{row_elapsed}` (custom keys can't use the template `:.dim` modifier), pointing at the same `reporter/CLAUDE.md` §8 note as `update_summary_msg`.
- **Minor (duplicate digit_count):** added cross-reference comments on both twins (u64 in the sink, usize in mod.rs).
- **Minor (initial `0/0 running`):** seeded the placeholder with the real worker count.

**Not changed — `rows` mutex poison handling:** left as the existing `if let Ok` / `let … else` pattern. Recovering a poisoned `rows` lock (à la `state_lock`) would risk acting on inconsistent row state, and a poisoning only happens if a rayon worker panicked while holding it — already caught upstream by `catch_unwind` — so skipping the update is the safer benign behavior. Happy to add a clarifying comment if you'd prefer the intent spelled out.

The one CI failure on the prior push was `hooks/bg-output-silent.yml` (a "best-effort log removal" assertion in the hook/job subsystem, unrelated to this display-only change — it passed in the other 3 integration matrix cells). Fresh CI running on the new push.

PR #602, opened 2026-05-30
chore: harden fresh-clone mise bootstrap

avihut:
## Problem

A fresh `daft clone` on a clean machine failed during hook bootstrap with two compounding issues:

1. **`post-clone` never trusted the mise config.** The `mise-install` job ran `mise install` directly, so mise refused to parse the just-cloned `mise.toml`:
   > `Config files in ~/Projects/daft/master/mise.toml are not trusted.`

   The whole `post-clone` install chain (`mise-install` → `install-lefthook-in-repo`) died, so nothing was pre-warmed. `worktree-post-create` already trusts first; `post-clone` did not — a pure authoring asymmetry.

2. **`cargo:` tools compiled from source and blew the 300s hook timeout.** With nothing pre-installed, `worktree-post-create`'s `mise install` became the first cold install. `cargo-deny` + `cargo-cooldown` were built with `cargo install` (source compile), serializing on the `~/.cargo` package-cache lock, and the job was killed at exactly `5m 0s` (the default hook timeout). The machine had no `cargo-binstall` on PATH, so mise silently fell back to compiling — the original machine was fast only because it happened to have `cargo-binstall` installed.

## Fix

- **`daft.yml`**: add a `mise-trust` job to `post-clone` before `mise-install`, mirroring `worktree-post-create`.
- **`mise.toml`**: add `cargo-binstall` (pinned to the prebuilt `aqua` backend) to `[tools]` so the `cargo:` tools install from prebuilt GitHub-release binaries. mise's `cargo.binstall` defaults to `true` but only engages when `cargo-binstall` is on PATH, and mise does not auto-install it.
- **`mise.lock`**: regenerated to pin `cargo-binstall@1.18.1` (clears the 7-day cooldown gate).

`mise install` on the affected tools now fetches prebuilt binaries in seconds instead of a >5min compile, keeping both hooks well under the timeout. `daft hooks validate` passes.


avihut:
Thanks for the review. Addressing the one required action — the missing regression test.

I looked into this and concluded the regression-test rule doesn't have a faithful target here, for two reasons:

1. **No new code path.** The fix doesn't change daft's hook engine — it adds a missing job to *this repo's* `daft.yml`. The behavior it relies on (a job whose `needs` dependency was *skipped* still runs, because skipped needs are stripped from the DAG) is already covered by `kept_spec_needs_are_stripped_of_skipped_dependencies` in `src/hooks/job_adapter.rs:616`. That's precisely what lets `mise-trust` run when `install-mise` is skipped.

2. **A "fails-without-the-fix" test is structurally impossible.** The bug was an authoring omission in this repo's own `daft.yml`. Per Critical Rule #2, the YAML scenarios run against isolated temp repos with their *own* `daft.yml` fixtures — they can't reference or assert against this repo's config. A scenario defining its own `post-clone: [mise-trust, mise-install]` fixture would pass identically before and after this PR (it's testing the engine, which already works), so it wouldn't be a regression test for this change — it'd be a duplicate of the existing unit test.

The `cargo-binstall` half is a build-time performance guard, which (as you note) has no YAML-scenario surface.

Happy to add an engine-level scenario anyway if you'd prefer the belt-and-suspenders coverage, but flagging that it wouldn't actually gate the bug this PR fixes.

PR #607, opened 2026-06-01
fix(ci): avoid broken pipe in MSRV staleness version fetch

### LINKED ISSUE #606: msrv-staleness-check workflow fails: broken pipe in curl | awk version fetch (exit 23)
## Symptom

The scheduled `MSRV staleness check` workflow (`.github/workflows/msrv-staleness-check.yml`) failed on its first run (2026-06-01, run [26742377175](https://github.com/avihut/daft/actions/runs/26742377175)). The "Compute MSRV gap and surface issue" step exited with:

```
curl: (23) Failure writing output to destination
##[error]Process completed with exit code 23.
```

No "Declared MSRV / Gap" line was printed — the step died before computing the gap, so no tracking issue was ever opened.

## Root cause

The version-fetch pipeline pipes a large file straight into an `awk` that `exit`s on the first match:

```bash
LATEST=$(curl -fsSL .../channel-rust-stable.toml \
  | awk '... f && /^version *=/{ ...; print ...; exit }')
```

`channel-rust-stable.toml` is ~858 KB. `awk` matches the `[pkg.rust]` version near the top and calls `exit`, closing the read end of the pipe while curl still has most of the file to write. curl gets `EPIPE` and exits with its write-output failure code (**23** on the runner; some curl builds report the same message as **56**). Because the step runs under `set -euo pipefail`, `pipefail` propagates curl's non-zero status and `set -e` aborts the step — even though the parse itself succeeded.

It surfaced now because the workflow was added in #481 and the monthly `0 6 1 * *` cron had its first execution on 2026-06-01. The bug has been latent since it was written. (It is also bash-version-sensitive: modern bash aborts a failed command-substitution-in-assignment under `set -e`; older bash swallows it, so a naive local repro can look fine.)

## Fix

Decouple the fetch from the parse so nothing closes the pipe early (and `curl -f`'s HTTP-error handling stays unambiguous):

```bash
curl -fsSL .../channel-rust-stable.toml -o channel-rust-stable.toml
LATEST=$(awk '...' channel-rust-stable.toml)
```

## Note (separate, real signal)

The gap the check was about to report is genuine: declared MSRV **1.93** vs current stable **1.96.0** = **3** minor versions, exceeding the `latest_stable - 2` policy. Once this fix lands (or the workflow is dispatched manually), the workflow will correctly open a "bump MSRV 1.93 → 1.94" tracking issue. That bump is separate human-triage work and is out of scope for this fix.

avihut:
## What

The `MSRV staleness check` workflow failed on its first scheduled run (2026-06-01, [run 26742377175](https://github.com/avihut/daft/actions/runs/26742377175)) with `curl: (23) Failure writing output to destination`, aborting before it computed the MSRV gap.

## Why it failed

The version fetch piped a ~850 KB file (`channel-rust-stable.toml`) straight into an `awk` that `exit`s on the first match. `awk` closes the read end of the pipe early; curl can't finish writing and exits **23**, and `set -euo pipefail` propagates that as the step's failure — even though the parse had already succeeded. Latent since the workflow was added in #481; it only surfaced now because the monthly `0 6 1 * *` cron had its first execution.

## Fix

Download the channel file first, then parse it from disk — there's no pipe to break. `curl -f` still surfaces a genuine HTTP error cleanly.

## Verification

Ran the fetch/parse/gap logic under **bash 5.x** (runner-accurate — older bash masks the failure) with `set -euo pipefail`:

- exits **0**
- computes `Declared MSRV: 1.93  Current stable: 1.96.0  Gap: 3`

A conventional unit/YAML regression test doesn't map to a GitHub Actions shell step (the repo has no test harness for workflows). The workflow keeps its `workflow_dispatch` trigger, so it can be validated end-to-end by dispatching it manually.

## Note (separate work)

The gap it now computes (3, vs the policy's 2) is real: once this merges and the workflow runs, it will open a "bump MSRV 1.93 → 1.94" tracking issue. That bump is separate human-triage work and is intentionally out of scope here.

Fixes #606

### DIFF
--- a/.github/workflows/msrv-staleness-check.yml
+++ b/.github/workflows/msrv-staleness-check.yml
@@ -52,17 +52,20 @@ jobs:
           # ~850 KB file to write; curl then dies with a write-output error
           # (exit 23) that `pipefail` propagates, aborting the step even though
           # the parse already succeeded. Downloading first removes the pipe
-          # (and keeps `curl -f`'s HTTP-error handling unambiguous).
+          # (and keeps `curl -f`'s HTTP-error handling unambiguous). $RUNNER_TEMP
+          # is the runner's per-job scratch dir (emptied each job), so the file
+          # never lands in the checked-out workspace.
           #
           # The `[pkg.rust]` block's `version` field carries the form
           #   version = "1.96.0 (ac68faa20 2026-05-25)"
           # — extract the leading semver triple.
+          channel_toml="$RUNNER_TEMP/channel-rust-stable.toml"
           curl -fsSL https://static.rust-lang.org/dist/channel-rust-stable.toml \
-            -o channel-rust-stable.toml
+            -o "$channel_toml"
           LATEST=$(awk '/^\[pkg\.rust\]$/{f=1; next} f && /^version *=/{
               match($0, /[0-9]+\.[0-9]+\.[0-9]+/);
               print substr($0, RSTART, RLENGTH); exit
-            }' channel-rust-stable.toml)
+            }' "$channel_toml")
 
           if [ -z "$CURRENT" ] || [ -z "$LATEST" ]; then
             echo "::error::Failed to parse MSRV (current=$CURRENT, latest=$LATEST)"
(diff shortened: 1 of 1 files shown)


avihut:
Thanks for the review. Addressed the one nit in 9ba73c96 — the channel manifest now downloads to `$RUNNER_TEMP/channel-rust-stable.toml` instead of the workspace root, so it never touches the checked-out tree. I went with `$RUNNER_TEMP` over `/tmp` since it's GitHub's documented per-job scratch dir (emptied at the start/end of each job, guaranteed writable). Re-verified under bash 5.x: exits 0, computes the gap (1.93 vs 1.96.0 = 3), and the full 858 KB downloads with no broken pipe.

PR #617, opened 2026-06-02
feat: DAG-aware --skip-hooks flag for worktree hooks (replaces --no-hooks)

### LINKED ISSUE #616: `--skip-hooks`: per-invocation skip/exclude of worktree-creation hooks (DAG-aware, attributed)
## Motivation

Creating a worktree (`worktree-checkout`, `worktree-checkout-branch`, `start`) always
fires the configured hooks when the repo is trusted. There's currently no per-invocation
way to:

- create a worktree **without** running hooks — only `clone` and `flow-adopt` expose
  `--no-hooks` (`src/commands/clone.rs:109`, `src/commands/flow_adopt.rs:106`); the
  checkout commands have no equivalent; or
- run **all but a few** jobs (e.g. skip the heavy `build` / network jobs when spinning up
  a quick scratch worktree).

`daft.yml` can gate jobs statically with `skip:` / `only:` / `exclude_tags`, but those are
config, not a runtime lever. This adds a single runtime flag that covers both needs.

## Proposed flag

One flag on the worktree-creation commands, value-driven:

```
--skip-hooks <SELECTOR>[,<SELECTOR>...]        (repeatable)
```

| Selector        | Meaning                                                        |
| --------------- | -------------------------------------------------------------- |
| `all` (or `*`)  | every job — subsumes the old `--no-hooks`                      |
| `tag:<tag>`     | all jobs carrying `<tag>`, **plus their dependents**           |
| `<name>` (bare) | the job named `<name>`, **plus its dependents**                |
| `job:<name>`    | explicit name form — escape hatch for a job named `all`/`*` or containing `:` |

```
daft start feat/x --skip-hooks all                 # skip everything
daft start feat/x --skip-hooks lint                 # drop lint (+ its dependents)
daft start feat/x --skip-hooks tag:heavy            # drop all heavy jobs (+ dependents)
daft start feat/x --skip-hooks tag:heavy,lint       # mix tags and names
daft start feat/x --skip-hooks tag:heavy --skip-hooks lint   # repeatable, same effect
```

Bare token = **job name**; tags require the `tag:` prefix. The asymmetry is deliberate —
it keeps the no-match warning honest (you declared which namespace you meant).

**clap shape** (the mandatory value is load-bearing):

```rust
/// Skip hooks this run. Repeatable / comma-separated.
/// Selectors: `all`, `tag:<tag>`, or a job name. See daft-hooks(1).
#[arg(long, value_name = "SELECTOR", value_delimiter = ',', num_args = 1..)]
skip_hooks: Vec<String>,
```

`num_args = 1..` forces at least one selector, so `--skip-hooks all feat/x` is unambiguous
(`all` is the value, `feat/x` the positional branch). A bare/optional-value flag would
swallow the positional branch arg on `start`/`checkout` — avoid it.

**Replace `--no-hooks`** on `clone`/`flow-adopt` with `--skip-hooks all` for a uniform
surface across every hook-firing command. We're effectively pre-1.0, so cut cleanly rather
than carry two spellings (optionally retain `--no-hooks` purely as a typing alias for
`--skip-hooks all`).

## Semantics: downstream cascade (the crux)

This is **not** a `jobs.retain(...)` one-liner. `src/hooks/job_adapter.rs:159` currently
strips `needs:` references to filtered-out jobs and treats a removed dependency as
**"vacuously satisfied"**:

```rust
// Remove needs: references to skipped jobs so dependent jobs don't
// fail DAG construction. A skipped dependency is vacuously satisfied …
```

That's correct for `hooks run --job build` (re-run `build`; `install` already happened),
but it's the **exact inverse** of what `--skip-hooks` needs: if I deliberately skip
`install`, running `build` (which `needs: [install]`) against an absent install is the
broken outcome. So exclusion must remove the matched jobs **and the transitive closure of
jobs that `needs` them** (downstream dependents). Upstream deps are left untouched —
orphan-pruning saves time but never protects correctness and silently dropping a job you
didn't name is surprising (non-goal).

Worked example:

```yaml
install:                      # no needs
build:   { needs: [install], tags: [heavy] }
test:    { needs: [build] }
lint:    { needs: [install] }
```

- `--skip-hooks install` ⇒ drops **install, build, test, lint** (all reach install).
- `--skip-hooks tag:heavy` ⇒ drops **build** and **test** (test needs build); `install`
  and `lint` run.

Because dependents are removed up front, the adapter's `needs`-stripping at
`job_adapter.rs:159` becomes a no-op for this path — the two never fight, and it stays
correct for the existing `hooks run --job` include path.

## Attributed skip reporting (required)

Excluded jobs must be **shown as skipped with the reason** — not silently dropped — so the
hook output makes it obvious *why* a job didn't run. The plumbing already exists; this is
about routing exclusions into the reported-skip channel instead of the silent
`jobs.retain(...)` at `src/hooks/yaml_executor/mod.rs:214-246`:

- `SkippedJob { name, reason: String }` — `src/hooks/job_adapter.rs:21`
- `HookResult.skip_reason: Option<String>` — `src/hooks/executor.rs:32`
- `JobOutcome::Skipped { reason }` → `on_job_skipped` → `finish_job_skipped`
  (`src/executor/cli_presenter.rs:90`, `src/output/hook_progress/mod.rs:119`)

Reason attribution:

| Case                                  | Reason string (rendered next to the job)            |
| ------------------------------------- | --------------------------------------------------- |
| Direct selector match                 | `requested (--skip-hooks)`                          |
| Cascade-removed dependent             | `depends on <job> (skipped)` — mirrors `NodeStatus::DepFailed` ("skipped because a dependency failed", `src/executor/mod.rs:182`) |
| `--skip-hooks all` (whole hook)       | `HookResult::skipped("all hooks skipped by request")` (matches existing `skipped("Hooks are globally disabled")` etc.) |

## Implementation sketch

1. Extend `JobFilter` (`src/hooks/yaml_executor/mod.rs:70`) with the exclude side —
   `exclude_job_names: Vec<String>`, `exclude_tags: Vec<String>`, and an `all` flag
   (or model selectors as a small `enum Skip { All, Job(String), Tag(String) }`).
2. Parse `--skip-hooks` values into selectors → `JobFilter`; thread through the existing
   `HookExecutor::with_job_filter` (`src/hooks/executor.rs:211`). `all` short-circuits the
   whole fire (the old no-hooks path).
3. Compute the exclusion set + **reverse-`needs` closure** on the `Vec<JobDef>` (while
   `needs` is intact), at the current `exclude_tags` retain site (`mod.rs:214`) —
   **before** `yaml_jobs_to_specs` and **before** `partition_foreground_background`
   (so pruning happens prior to bg→fg promotion in `src/hooks/yaml_executor/partition.rs`).
4. Route the excluded `JobDef`s into `SkippedJob` entries with attributed reasons so the
   presenter and log store surface them (instead of dropping silently).
5. Leave `job_adapter.rs:159` as-is (correct for the include path; no-op for exclude).

## Decisions

- **Naming: `--skip-hooks`** (locked). "Skip" fits the union of behaviors now that the flag
  also covers `all`; `--exclude-*` was the alternative (aligns with the `exclude_tags`
  config term) but is narrower.
- **No-match selector ⇒ visible warning.** Empty-after-skip is a no-op (not an error,
  unlike include's `bail!("No job named …")`), but a typo'd `--skip-hooks tag:hevy` matching
  nothing means the job **runs** when you thought it was gone — silent no-match is more
  dangerous in the exclude direction, so warn per unmatched selector.
- **Group exclusion ⇒ whole subtree.** Skipping a job holding a nested `group:` skips its
  subtree. (Note: group jobs are currently emitted as `SkippedJob`s anyway —
  `job_adapter.rs` "group jobs are not yet supported by the generic executor" — so confirm
  interaction.)

## Affected areas

- **`area:hooks`** — `JobFilter`, the cascade, skip attribution (bulk of the work).
- **`area:commands`** — the flag on `worktree-checkout` / `worktree-checkout-branch` /
  `start`; replace `--no-hooks` on `clone` / `flow-adopt`.
- Lighter touches: completions (a value completer for `all` / `tag:` / job names —
  `src/commands/completions/{bash,zsh,fish}.rs`) and output (the skip-reason strings).

## Acceptance criteria / tests

- **Unit (cascade):** skip `install` ⇒ `build`/`test`/`lint` all removed; `tag:` selector;
  mixed name+tag; `all` short-circuits; group subtree.
- **Unit (attribution):** directly-matched jobs report `requested (--skip-hooks)`;
  cascade-removed dependents report `depends on <job> (skipped)`; both appear in results
  rather than being absent.
- **Unit (safety):** unmatched selector emits a warning and proceeds; empty-after-skip is a
  no-op, not an error.
- **YAML scenario** under `tests/manual/scenarios/` exercising `--skip-hooks` on a worktree
  create against a multi-job `daft.yml` (assert kept vs skipped sets and the rendered
  reasons).
- Regenerate man pages (`mise run man:gen`), update shell completions, and update
  `SKILL.md` + docs for the new flag (per CLAUDE.md).

avihut:
## Summary

Replaces the boolean `--no-hooks` with a value-driven, DAG-aware `--skip-hooks <SELECTOR>...` flag on every hook-firing command (`checkout`, `checkout-branch`, `go`, `start`, `clone`, `flow-adopt`). It gives a per-invocation lever to skip hooks wholesale or drop a subset of jobs **and everything that depends on them**, with each excluded job rendered as skipped (attributed reason) and recorded — never silently dropped.

Fixes #616

## Selector grammar

| Selector | Effect |
|---|---|
| `all` / `*` | skip the whole fire (the `--no-hooks` replacement; short-circuits, preserves trust gating) |
| `<hook>` (e.g. `worktree-post-create`) | skip that one hook wholesale; each job shown as skipped |
| `tag:<tag>` | skip tagged jobs **+ their reverse-`needs` dependents** |
| `<name>` (bare) | skip that job **+ its dependents** |
| `job:<name>` | escape hatch for a job literally named `all` / a hook-type |

Repeatable and comma-separated: `--skip-hooks tag:heavy,lint` or `--skip-hooks build --skip-hooks test`.

## Notable changes

- **`--no-hooks` removed** (no alias) — pre-1.0 clean cut; `--skip-hooks all` is the uniform replacement. Trust semantics preserved: `all` still gates config-load/trust; partial skips still imply trust on `--install`.
- **Cascade engine** (`compute_skip_cascade`): pure reverse-`needs` BFS, cycle-safe, deterministic immediate-parent attribution.
- **Surfaces**: shell completions (bash/zsh/fish) complete selector values from `daft.yml`; regenerated man pages; updated docs/SKILL; 2 new YAML scenarios + updated integration test.

## Test plan

- `mise run test:unit` — 1943 pass (cascade, parser, engine-render, clap-parse tests)
- `mise run clippy` / `mise run fmt:check` — clean
- `mise run test:manual hooks` — skip-hooks-cascade + skip-hooks-by-hook-name + flags-exclusive + no-hooks-flag all green

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -620,6 +620,36 @@ daft hooks run worktree-post-create --verbose    # Show skipped jobs with reason
 Use cases: re-running after a failure, iterating during hook development, or
 bootstrapping existing worktrees that predate the hooks config.
 
+### Skipping Hooks Per-Invocation (`--skip-hooks`)
+
+The worktree-creating commands (`daft start`, `daft go`,
+`git worktree-checkout`, `git worktree-checkout-branch`) and
+`git worktree-clone` / `git worktree-flow-adopt` accept `--skip-hooks` to
+exclude jobs for one run (repeatable / comma-separated):
+
+```bash
+daft start feat/x --skip-hooks all          # skip every hook (replaces the old --no-hooks)
+daft start feat/x --skip-hooks lint          # skip the lint job AND its dependents
+daft start feat/x --skip-hooks tag:heavy     # skip every heavy-tagged job AND dependents
+daft start feat/x --skip-hooks tag:heavy,lint
+git worktree-clone <url> --skip-hooks all    # clone without running any hooks
+```
+
+Selectors: `all` / `*` (every job), `tag:<tag>` (tagged jobs + dependents),
+`<name>` (a job + its dependents), `job:<name>` (explicit-name escape hatch). A
+bare token is a job **name**; tags need the `tag:` prefix.
+
+Key behavior — the **downstream cascade**: skipping a job also skips every job
+that `needs:` it (transitively), because running a dependent against a
+deliberately-skipped dependency is broken. Upstream dependencies are untouched.
+Excluded jobs are reported as skipped with a reason, not dropped silently; a
+selector matching nothing warns and the run proceeds.
+
+`--skip-hooks` is the **exclusion** counterpart to `daft hooks run --job/--tag`
+(which is an inclusion filter). `--skip-hooks all` cannot be combined with
+`--trust-hooks`; a partial skip (`tag:`/`<job>`) still runs your own hooks and
+remains compatible with `--trust-hooks`.
+
 ### Move Hooks
 
 When a worktree is moved (rename via `daft worktree-branch -m`, layout transform
--- a/SKILL.md
+++ b/SKILL.md
@@ -629,15 +629,20 @@ exclude jobs for one run (repeatable / comma-separated):
 
 ```bash
 daft start feat/x --skip-hooks all          # skip every hook (replaces the old --no-hooks)
+daft start feat/x --skip-hooks worktree-post-create  # skip one whole hook by name
 daft start feat/x --skip-hooks lint          # skip the lint job AND its dependents
 daft start feat/x --skip-hooks tag:heavy     # skip every heavy-tagged job AND dependents
 daft start feat/x --skip-hooks tag:heavy,lint
 git worktree-clone <url> --skip-hooks all    # clone without running any hooks
+git worktree-clone <url> --skip-hooks post-clone  # clone, run worktree hooks but not post-clone
 ```
 
-Selectors: `all` / `*` (every job), `tag:<tag>` (tagged jobs + dependents),
-`<name>` (a job + its dependents), `job:<name>` (explicit-name escape hatch). A
-bare token is a job **name**; tags need the `tag:` prefix.
+Selectors: `all` / `*` (every job), `<hook>` (a whole hook by its canonical
+`daft.yml` key, e.g. `worktree-post-create` / `post-clone`), `tag:<tag>` (tagged
+jobs + dependents), `<name>` (a job + its dependents), `job:<name>`
+(explicit-name escape hatch). A bare token resolves in order: wildcard → hook
+type → job name; tags need the `tag:` prefix. A hook-type selector that names a
+hook the command never fires is a silent no-op (no error, no warning).
 
 Key behavior — the **downstream cascade**: skipping a job also skips every job
 that `needs:` it (transitively), because running a dependent against a
--- a/docs/cli/daft-go.md
+++ b/docs/cli/daft-go.md
@@ -54,6 +54,7 @@ Cannot be combined with `-b`/`--create-branch`.
 |--------|-------------|---------|
 | `-s, --start` | Create a new worktree if the branch does not exist | |
 | `--local` | Skip all remote operations (no fetch) for this invocation | |
+| `--skip-hooks <SELECTOR>` | Skip hooks when `go` creates a worktree (`all` \| a hook name like `worktree-post-create` \| `tag:<tag>` \| `<job>`); repeatable/comma-separated | |
 | `-x, --exec <EXEC>` | Run a command in the worktree after setup (repeatable) | |
 | `--no-cd` | Do not change directory to the new worktree | |
 | `-c, --carry` | Apply uncommitted changes from the current worktree | |
--- a/docs/cli/daft-start.md
+++ b/docs/cli/daft-start.md
@@ -39,6 +39,7 @@ operations for a single invocation regardless of config.
 | Option | Description | Default |
 |--------|-------------|---------|
 | `--local` | Skip all remote operations (no fetch, no push) for this invocation | |
+| `--skip-hooks <SELECTOR>` | Skip hooks this run (`all` \| a hook name like `worktree-post-create` \| `tag:<tag>` \| `<job>`); repeatable/comma-separated | |
 | `-c, --carry` | Apply uncommitted changes from the current worktree to the new one | |
 | `--no-carry` | Do not carry uncommitted changes | |
 | `-x, --exec <EXEC>` | Run a command in the worktree after setup completes (repeatable) | |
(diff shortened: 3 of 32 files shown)


avihut:
Thanks for the review. Triage and actions (pushed in ecf6b404):

**Addressed**
- **BFS cycle-safety (the one you flagged for before-landing):** added `cascade_terminates_on_cyclic_needs` — proves `compute_skip_cascade` terminates and attributes correctly on `A needs B, B needs A`, backing the "cycle-safe" doc claim that was previously only design-asserted.
- **`complete_skip_hooks` union scope:** added a comment documenting that selector-value completion intentionally unions jobs/tags across *all* configured hooks (per-command firing-set derivation is repo-state dependent; an offered-but-non-firing selector just warns).

**Deferred (good follow-ups, not folded into this PR to keep it tight)**
- *Skipped-job `JobRow` persisted with empty tags* — real gap, but only matters once `daft hooks jobs` grows tag-based filtering; needs a `tags` field threaded through `SkippedJob`.
- *Completion `job:` escape for shadowed names* — niche (a job literally named `all`/a hook-type); the escape hatch works when typed, and the better fix is the completion-side change you suggested.
- *`skip_hooks_all` re-parse* and *two-copy `requested_skips`* — both confirmed benign; tidy-ups, not correctness.
- *Checkout-path `--skip-hooks all` scenario* — engine is unit-tested and the clone path is integration-tested; a parallel checkout scenario would close the last loop.

Note: the one red CI cell on the prior push was `integration-tests (default, yaml)` failing on `hooks/bg-output-silent.yml` step 4/6 — a known matrix-hopping flake (best-effort log-removal race), unrelated to this change. Fresh CI is running on ecf6b404.

PR #629, opened 2026-06-13
fix: provenance-gated visitor daft-file propagation

### LINKED ISSUE #628: Visitor daft.yml propagation overwrites the default-branch config with stale copies on merge/removal
## Problem

The visitor-config propagation treats the worktree being merged/removed as the authoritative copy of the untracked daft files. Because every sync primitive is a **two-way source-wins merge**, a stale inherited copy silently overwrites the canonical config.

### Reproduction (field report)

1. `main`'s untracked visitor `daft.yml` contains version **A**.
2. `git-worktree-checkout -b feat/x` seeds A into the new worktree (correct).
3. `main`'s `daft.yml` evolves to version **B**. `feat/x`'s copy is never touched — it is a pristine, stale snapshot of A.
4. Delete the branch (`daft merge feat/x`, merge cleanup, or `daft remove -f feat/x`).
5. `main/daft.yml` reverts to A — silently, with no output. The file is untracked, so git offers no recovery.

### Mechanism

- `propagate()` is source-wins and rewrites the target (`src/hooks/visitor_propagation.rs:31-100`).
- The removal "salvage" block runs it silently — `let _ = propagate(...)` (`src/core/worktree/branch_delete.rs:929-964`).
- `daft merge` overlays the source's files onto the target before the git merge (`src/commands/merge.rs:696`), visible only at debug level.
- Merge cleanup hardwires `force: true` (`src/commands/merge.rs:1035`) to skip a reachability check, which also bypasses the divergence guard.
- The two-way divergence check cannot distinguish "the worktree refined its copy" from "the target moved on" — so the unforced guard also **false-positively refuses** removal of pristine-but-stale worktrees.
- The refusal message says force will "remove anyway", but force actually propagates; `prune --force` discards. Same operation, three semantics.

### Related gap

`prune`/`sync` assume remote-gone ⇒ merged: gone branches are removed and the local branch is **force**-deleted (`src/core/worktree/prune.rs:856-865`) with no `is_branch_merged` check (which exists in branch_delete validation but is not wired into prune). An abandoned (never merged) branch whose remote was deleted is destroyed without warning.

## Agreed behavior

Provenance-gated propagation: record the content daft seeds into each worktree (SQLite store); classify copies as **Pristine** (== seed) / **Refined** (edited since seed) / **NoSeed**; all consolidation is a **three-way merge** (base = seed) where conflicting keys are never auto-resolved.

| Command | Pristine | Refined | Conflict |
|---|---|---|---|
| checkout / checkout-branch | copy verbatim + record seed | — | — |
| daft merge | skip file (target untouched) | three-way into target, announce keys taken | per-key prompt; non-interactive abort before the git merge |
| remove / branch-delete (unforced) | delete silently | prompt to consolidate (interactive) / refuse pointing at `daft file merge` | same |
| forced (-f / -D) | delete | discard + stash to `.daft/discarded/`; never write the target | same |
| prune / sync | remove silently | skip + end-of-run summary | same |
| prune --force | remove | discard + stash + summary | same |
| daft file merge | "nothing to do" | three-way with preview, backup, seed update | per-key prompt; `-y` = source-wins |

Plus: prune/sync gain the `is_branch_merged` guard — gone-but-unmerged branches are skipped with a warning unless `--force`.

Invariants: the target's files are only ever written by `daft merge` (announced) or `daft file merge` (explicit); force always means discard, never propagate; every cross-worktree write is printed and backed up; store failures degrade to NoSeed semantics and never block worktree operations.

avihut:
## Problem

Deleting a worktree (or running `daft merge` / prune) could silently overwrite the default branch's untracked visitor `daft.yml` with a stale seeded copy. Concretely: master's `daft.yml` evolves A -> B; a feature worktree still holds the seeded copy A; on `daft merge`, forced `daft remove`, or prune cleanup, daft overlaid A back onto B — reverting the canonical config with no prompt and no output.

Root cause: two-way **source-wins** propagation could not tell a *refined* copy from a *stale-but-untouched* one. The removal salvage block, merge-time `propagate_atomic`, and merge cleanup's hardwired `force: true` all wrote the source over the target unconditionally. The unforced guard also *false-positively refused* removing pristine-but-stale worktrees, and prune/sync assumed "remote gone => merged", force-deleting unmerged branches.

## Fix: seed provenance

daft now records the exact content it seeds into each worktree's untracked daft files (`visitor_seeds` table, migration 002) and classifies every copy against its seed — **Pristine** (== seed), **Refined** (edited), or **NoSeed** — at every lifecycle point:

- **Pristine / subsumed copies delete silently** — forced/unforced remove, merge, and prune all leave the target untouched (kills the A -> B -> A revert and the old false-positive refusal).
- **Refined copies are never silently lost or merged** — removal prompts consolidate / discard / abort with a key-level summary; scripts refuse and point at `daft file merge`; force *discards* to `.git/.daft/discarded/<branch>/` instead of writing the target.
- **`daft merge` consolidates three-way** (seed as base), announces the adopted keys, aborts before the git merge on conflicting keys, and refreshes the source seed so cleanup stays silent.
- **Prune / sync** verify `is_branch_merged` before destroying gone branches (gone-but-unmerged is kept with a `--force` hint) and never write another worktree.
- **`daft file merge`** gained three-way mode with a preview, a pre-write backup, per-key conflict prompts (`-y` = source-wins), and seed bookkeeping.

## Tests

RED-first: 12 manual scenarios reproducing each failure landed in the first commit and failed on the old binary for the predicted reasons. Final state is green across `mise run ci` — 632 manual scenarios, both bash integration matrix cells (default + gitoxide), unit tests, and man/CLI-docs verify. Docs, SKILL.md, glossary/FAQ/recipe, help text, and man pages are updated.

Fixes #628


avihut:
Addressed the review — triage below.

## Fixed

**[Bug] Default branch destroyed on prune** (`identify_gone_branches`, commit `cb67388f`)

Confirmed real and reproduced it: an isolated `trunk`-default repo with a deleted upstream had its **default-branch worktree _and_ local branch** pruned (`* — pruned trunk — removed worktree, local branch`). It was slightly broader than flagged — Method 2 name-excluded only `master`/`main`, Method 1 (`: gone]`) excluded the default branch at all, and the gone-but-unmerged guard deliberately *proceeds* on the default branch (trivially merged into itself). The resolved default branch is now excluded **by identity in both identification methods**, keeping `master`/`main` only as the fallback when the default can't be resolved (so they stay protected under `--force` with an unresolvable default). All four callers pass the resolved default.

Added a bash regression test (`prune_keeps_default_branch_with_nonstandard_name`) — RED before, GREEN after — which also closes the "no test for Method 2 with a non-standard default branch" gap the review called out.

**Duplicate `SeedsContext` opens / silent rollback** (commit `7e8f4055`)

- `remove_worktree` and `plan_refined_files` each opened a second store handle; both now reuse the one already in scope.
- `write_files_atomic` swallowed failed rollback writes; they now `log_debug!`, matching the best-effort logging in `visitor_seeds`.

## Left as-is (the review itself concluded these are correct)

- **`merge3_jobs` unnamed-job adoption** — intentional conservative behavior (no identity to diff an unnamed job against); already documented in-code, and flagged as "not a bug."
- **`merge3` `log: {}` → `None` collapse** — correct given `#[serde(default)]` + `Option` semantics; there is no "absent" vs "explicitly default" distinction, as the review reasoned.

## Note

Merged latest master (dep/workflow bumps + the #635 `bg-output-silent` de-flake). The one conflict resolved to keep #635's bounded poll loop together with this branch's `DAFT_STATE_DIR`-aware job path; ran that scenario 3× clean.

CI is green across the full matrix — build, lint, msrv, windows, xtask, homebrew, and all four `integration-tests` cells (default/gitoxide × bash/yaml).

Issue #600, opened 2026-05-30
feat(push): daft push — run pre-push hooks in the pushed branch's worktree


avihut:
## Problem

daft has no `push` command, and plain `git push <branch>` runs the `pre-push` hook with **cwd = the worktree you invoked push from**, not the worktree that owns the branch being pushed. Native `pre-push` hooks are shared across all worktrees of a bare repo (resolved from `$GIT_COMMON_DIR/hooks` or `core.hooksPath`), and git fires the hook exactly once with the invoking worktree's cwd.

Live probe (bare `origin.git` + worktrees `featureA`/`featureB` + a shared `pre-push` that echoes its cwd): pushing `featureB` *from featureA's worktree* gives the hook `cwd = .../featureA`, `GIT_DIR = .../worktrees/featureA`, toplevel `.../featureA` — the WRONG tree. Pushing from featureB's own worktree gives `.../featureB` (correct). A `pre-push` hook that runs the test suite, lints the working tree, or reads worktree-local config (`.env`, `mise`, `direnv`) is therefore silently run against the wrong source while pushing a different branch.

`daft push <branch>` exists to fix exactly this: run the push (and thus the shared `pre-push` hook) with cwd set to the **pushed branch's** worktree. That worktree-correct hook context is its entire raison d'être.

## Scope (intentionally tiny)

1. Resolve `<branch>` → its worktree path via `GitCommand::find_worktree_for_branch` (`src/git/worktree.rs:124`, returns `Result<Option<PathBuf>>`).
2. Push via the shared seam `push_with_hooks` (`src/core/worktree/push.rs`, built by #599) with `worktree_cwd = Some(that path)`, `verify=true` (do NOT pass `--no-verify`).
3. Honor `pre-push` and report hook runs/results through the seam's `JobPresenter` path: Path A synthetic single-job for native/lefthook/husky/pre-commit (NOW); Path B per-job when #468 manages the stage (LATER).
4. Wire into the shell wrapper's cd-redirect path? **No** — `daft push` does not change the filesystem layout the user's cwd lives inside, so it does NOT need a `DAFT_CD_FILE` arm.

## Non-goals (guard against scope creep)

[...]

Issue #468, opened 2026-05-03
feat(hooks): full git-hooks drop-in (lefthook replacement) covering commit-stage and beyond


avihut:
## Goal

Extend daft's hooks system from worktree-lifecycle-only to a **full git-hooks drop-in** — a lefthook-style replacement that fires on the standard git hook stages (\`pre-commit\`, \`commit-msg\`, \`pre-push\`, \`prepare-commit-msg\`, etc.) in addition to the existing worktree-lifecycle hooks.

This unifies feature work and docs in a single tracking issue (per "docs and features must enter together"). Supersedes the docs-only #465.

## Why

The Hooks pillar's positioning thesis (per the docs IA spec at \`docs/superpowers/specs/2026-05-03-docs-ia-restructure-design.md\`) is **"daft hooks as a local parallel to GitHub Actions, with each code-evolution stage getting its own gate."**

Today daft hooks cover only the worktree-lifecycle stage. The lefthook drop-in expands coverage to **commit-stage hooks** — the "progressive code-replication boundary" where format/lint/fast-tests run before a change is recorded. Together with merge hooks (#330) and worktree hooks (shipped), this completes the boundaries-as-stages picture.

It also opens a clear migration path for users on lefthook today, broadening daft's adoption surface beyond the worktree-only audience.

## Feature scope

- [ ] **Hook stages supported** (initial): \`pre-commit\`, \`commit-msg\`, \`prepare-commit-msg\`, \`pre-push\`, \`post-commit\`, \`pre-rebase\`. Beyond these as future work.
- [ ] **Implementation strategy**: register daft as the git hook handler (write thin shim scripts under \`.git/hooks/\` that delegate to daft); preserve any existing user-managed hooks (don't clobber)
- [ ] **\`daft.yml\` schema extension**: new top-level keys for commit-stage hook stages, sharing the existing job-orchestration semantics (parallelism, deps, conditions, OS/arch gating, trust)
- [ ] **Stage-specific context env vars**: \`pre-commit\` exposes staged files; \`commit-msg\` exposes message path; \`pre-push\` exposes ref ranges; etc.
- [ ] **Trust model**: same trust-on-first-use model as existing hooks (already in place)
- [ ] **Bypass**: \`--no-verify\` honored (git's standard); per-hook \`skip:\` matchers
- [ ] **Idempotent install**: re-running registration is a no-op; uninstall command restores original \`.git/hooks\` state
[...]

PR #630, opened 2026-06-13
feat(push): honor and report pre-push hooks at every daft push site

### LINKED ISSUE #599: feat(push): honor + report pre-push hooks at every daft push site (drop unconditional --no-verify)
## Problem: daft silently bypasses every pre-push hook

Every daft-initiated push hardcodes `--no-verify`, unconditionally bypassing the repo's `pre-push` hook — native `.git/hooks`, lefthook, husky, pre-commit-framework, all of it. The four primitives in `src/git/remote.rs` (verified bare `Command::new("git")`, `--no-verify` at lines 51/71/111/132) and a hand-rolled bypass:

- `push_set_upstream` — `remote.rs:49` (NO cwd — runs in process cwd)
- `push_set_upstream_from` — `remote.rs:64`
- `push_from` — `remote.rs:103` (returns captured stderr; load-bearing, see below)
- `push_delete` — `remote.rs:130` (NO cwd)
- **raw bypass** — `multi_remote.rs:747-761` `delete_remote_branch` builds `Command::new("git").args(["push", "--no-verify", remote, "--delete", branch])` directly, NOT through any primitive (it even takes an unused `_git`). A wrapper added only at the primitive layer silently misses this site.

The **6 primitive call sites**:

1. `sync --push` → `src/core/worktree/push.rs:153` (`push_single_worktree` → `git.push_from`)
2. worktree/branch create auto-upstream → `src/core/worktree/checkout_branch.rs:448` (`push_if_enabled`, called at `:182`)
3. worktree rename, push new name → `src/core/worktree/rename.rs:263` (`push_set_upstream_from`)
4. worktree rename, delete old remote branch → `src/core/worktree/rename.rs:270` (`push_delete`)
5. branch-delete, delete remote branch → `src/core/worktree/branch_delete.rs:855` (`push_delete`)
6. multi-remote move `--push` → `src/commands/multi_remote.rs:674` (`push_set_upstream`, wrapped in a non-thread-safe `set_current_dir`/restore dance at `:671-676`)

Result: a user with a `pre-push` test-gate or secret-scanner gets it run on `git push` but **silently skipped** on every daft push. That is wrong, and it's why daft can't yet claim to be a drop-in for hook-managed repos.

## Requirement 1 — Honor pre-push (NOW for native+lefthook+husky+pre-commit; later for daft hooks via #468)

Honoring is nearly free: dropping `--no-verify` makes git itself run `pre-push` and gate the push on its exit code. daft assembles **nothing** — git feeds the hook its own stdin (`<local-ref> <local-oid> <remote-ref> <remote-oid>` per ref). The existing `if !status.success() { bail!(stderr) }` in every primitive (verified `remote.rs:55,76,122,142`) already propagates a hook rejection as a daft error. So the NOW path (engine B: git dispatches lefthook/husky/native/pre-commit) = drop `--no-verify` + thread the correct worktree cwd. `HookExecutor` is NOT in this path — daft cannot re-run lefthook's pre-push as daft.yml jobs; only git can fire it. When #468 ships, daft additionally runs its own `pre-push` stage (Path B) for repos it manages.

## Requirement 2 — Report hook runs + results, mirroring remove/start/sync-removed-worktree

daft already reports lifecycle-hook runs in `worktree-post-create` (checkout/start), `worktree-pre/post-remove` (remove), and sync-for-removed-worktrees (prune). #599 mirrors that **reporting surface** for push hooks, reusing the same `JobPresenter` trait + renderers — but via TWO different drive mechanisms depending on path, because a native/lefthook `pre-push` is an opaque git subprocess, not a daft job:

- **Path A (the NOW path: native / lefthook / husky / pre-commit)** — drive the presenter adapters **DIRECTLY** around the streamed/teed git-push subprocess: `CliPresenter` (`src/executor/cli_presenter.rs:17`) / `TuiPresenter` (`src/output/tui/presenter.rs:74`) + `HookRenderer` (`src/output/hook_progress/mod.rs:14`), emitting a **single synthetic phase + job** (`on_phase_start("pre-push", Some(branch))` → one `on_job_*` carrying git's teed output → `on_phase_complete`). This does NOT route through `HookRunner::run_hook` / `CommandBridge` / `HookExecutor` / the `prune.rs:778` `run_removal_hook` mirror — those fire daft's OWN jobs through `HookExecutor::execute` (`progress.rs:111`), keyed on the **closed** `HookType` enum (`src/hooks/mod.rs:68` — variants PostClone/PreCreate/PostCreate/PreRemove/PostRemove/PreMerge/PostMerge; **no PrePush**), and return only coarse `HookOutcome`. An opaque native/lefthook subprocess has no `HookContext`, no `HookType`, no daft job, so it CANNOT go through that machinery. **This is not a parallel reporter** — same `JobPresenter` trait, same `HookRenderer` — it simply drives them directly rather than via `HookExecutor`.
- **Path B (LATER: daft owns the stage, #468)** — daft fires its own `pre-push` stage; THERE the `HookRunner::run_hook` / `run_removal_hook` (`prune.rs:778-807`, the canonical mirror — delegate rich reporting to the bridge's `run_hook`, keep a coarse `on_warning` fallback at `:798`) path applies, yielding per-job structured fidelity through the same presenter.

**Reporting is asymmetric — scope the promise honestly.** A native/lefthook `pre-push` is ONE opaque git subprocess; daft cannot see inside it. So Path A reporting is a **single synthetic phase/job**: phase header `pre-push` + captured output + overall pass/fail — NOT per-hook fidelity. Per-job structured reporting (hook name, skip reason) is achievable only on Path B. The issues must NOT promise per-hook reporting for lefthook.

**Same reporting TYPE (not just surface):** the single caller-facing return is `PushOutcome` (push status incl `up_to_date: bool` + a coarse hook verdict mirroring `HookOutcome` at `core/mod.rs:71`). Path B's rich `StageOutcome` is reported through the presenter for fidelity AND collapsed into `PushOutcome`'s coarse verdict; Path A's synthetic-job pass/fail collapses into the same coarse verdict. Callers branch on `PushOutcome` regardless of path.

## The up_to_date collision (a real design fork #599 must resolve)

`push_from` returns captured stderr and `push.rs:160-161` parses it for `"Everything up-to-date"` to set `up_to_date`. You cannot naively switch to `Stdio::inherit()` (the `pull_passthrough` pattern at `remote.rs:197`) to surface hook output — inheriting empties the captured string, so sync would mislabel up-to-date branches as "Pushed successfully." #599 must consciously pick **(a) tee** (capture for parsing AND echo for visibility) or **(b) restructure** detection (e.g. `git push --porcelain`). `run_push`'s post-condition makes this explicit: even under `PushStdio::Stream` it returns BOTH the exit-success bool AND the captured stderr, so the exit-code gate and the up_to_date parse both survive. Streamed/teed output is the *precondition* for the synthetic-job reporting above.

## Route through the shared seam — NOT 5 separate edits

All four primitives funnel through one private `run_push(push_args, cwd: Option<&Path>, verify, stdio) -> PushIo` in `src/git/remote.rs`; the composition lives in `push_with_hooks` in `src/core/worktree/push.rs` beside `push_single_worktree`. The 6 call sites keep calling the named primitives; only the 4 primitive bodies + the `multi_remote.rs:747` raw bypass (refactored to call `push_delete`) route through `run_push`. **cwd completeness**: `push_set_upstream` (`remote.rs:49`) and `push_delete` (`remote.rs:130`) run in process cwd today — they need cwd-bearing variants (or the seam threads cwd) so native hooks fire in the right worktree at sites 2/4/5; `multi_remote.rs:671-676`'s `set_current_dir` dance should migrate to a cwd-carrying primitive.

## The Path-B port: declared HERE, implemented by #468

#599 **declares** the `StageRunner` port (in the push path's ports module, consumer-owns-port per CLAUDE.md hexagonal rule) and **ships a Noop adapter whose `manages_stage` returns `false`** — the exact `NoopHookRunner` pattern at `core/mod.rs:93`. With the Noop, `manages_stage → false ⇒ Path A`, and **#599 compiles and ships fully on Path A with ZERO #468 code**. #468 later supplies the real adapter. The port:

```rust
trait StageRunner {
    fn manages_stage(&self, stage: &str, git_dir: &Path) -> bool;
    fn run_stage(&self, stage: &str, worktree_cwd: Option<&Path>,
                 refs: &[PushRef], presenter: Arc<dyn JobPresenter>) -> Result<StageOutcome>;
}
struct PushRef { local_ref, local_oid, remote_ref, remote_oid } // pins git's 4-field pre-push stdin
```

`push_with_hooks` takes `&dyn StageRunner` (NOT `&dyn HookRunner` — `run_hook(ctx: &HookContext) -> HookOutcome` at `core/mod.rs:85-88` cannot carry the pushed-branch `worktree_cwd` or the `PushRef`s and returns only coarse `HookOutcome`).

## Trust gating (the untrusted-repo rule — named, not left open)

Trust gates **only daft's OWN stage (Path B)**; it cannot gate git's native dispatch (Path A). On an untrusted repo (#596 Deny default), `manages_stage` effectively yields false → degrade to Path A → git still fires native/lefthook/husky/pre-commit → **the push proceeds**, daft's stage skipped, #596 emits the "config skipped" warning. (Nuance flagged for #468/#596, not resolved here: if #468 would have *chained* a foreign incumbent that daft owns via `core.hooksPath`, the untrusted-skip drops that chain too.)

## Opt-out

Since today's behavior is unconditional bypass, preserve an explicit `--no-verify` passthrough (`verify=false`) so callers who *want* to skip the hook can. The default flips from "always bypass" to "honor."

## Acceptance criteria

- A failing `pre-push` (native, lefthook, husky, or pre-commit) aborts each of the 6 sites + the former raw bypass with non-zero exit.
- `sync --push` still correctly detects "Everything up-to-date" after the stdio change (`run_push` returns stderr even under Stream; the collision is resolved, not regressed).
- Hook runs are reported through `JobPresenter` under both CLI and TUI (Path A: synthetic single-job driving `CliPresenter`/`TuiPresenter`+`HookRenderer` directly; Path B: per-job once #468 manages the stage).
- `--no-verify` passthrough still bypasses.
- The raw `multi_remote.rs:747` site goes through `push_delete`/`run_push` (greppable: no `"--no-verify"` literal survives outside the `run_push` verify toggle).

## Regression coverage

- YAML scenario(s) under `tests/manual/scenarios/` installing a `pre-push` that exits non-zero, asserting each push site aborts (and a passing hook lets it through).
- A test asserting `sync --push` up_to_date detection survives the tee/restructure change.
- A grep-gate test that `"--no-verify"` appears only inside `run_push` (no primitive or call site hardcodes it).

## Interfaces with #600 and #468

- **#600** consumes this seam; it adds only branch→worktree resolution (`find_worktree_for_branch`). The existing 6 sites are already cwd-correct, so #599 does NOT need resolution logic.
- **#468** implements the `StageRunner` port #599 declares. The seam calls `manages_stage("pre-push", git_dir)` → Path B (run daft stage via `run_stage`, report per-job, then `--no-verify`) when daft owns the stage; Path A (drop `--no-verify`, git dispatches) otherwise. #599 ships fully on Path A **today** against the Noop adapter without #468; #468 later upgrades the same report surface to structured fidelity. The seam must avoid double-invoking a foreign pre-push under Path B — #468 chains the incumbent inside its stage rather than letting git re-fire it.

avihut:
Fixes #599.

## What

Every daft-initiated push hardcoded `--no-verify`, silently bypassing the
repository's `pre-push` hook — native `.git/hooks`, lefthook, husky, and
pre-commit-framework alike. A developer with a pre-push test-gate or
secret-scanner got it on `git push` but never on a daft push, so daft could not
honestly claim to be a drop-in for hook-managed repos.

This flips the default from **always bypass** to **honor**: git now runs
`pre-push` and gates the push on its exit code at all six push sites — plus a
hand-rolled bypass that lived outside the primitives in `multi_remote.rs`. The
hook's run is reported through the same `JobPresenter` surface daft already uses
for lifecycle hooks, and `--no-verify` is preserved as an explicit
per-invocation opt-out.

It ships entirely on **Path A**: git dispatches the hook as one opaque
subprocess, daft tees its output and reports a single synthetic `pre-push`
phase. The `StageRunner` port that #468 will implement (daft-owned hook stages
with per-job fidelity) is **declared** here behind a Noop adapter, so this PR
carries zero #468 code while leaving the exact seam #468 plugs into.

### The shared seam (`src/git/remote.rs`)

- One private `run_push(args, cwd, opts) -> PushIo` funnels every push. The
  `--no-verify` literal now appears in exactly one place in the codebase, gated
  on `!opts.verify` — a `#[test]` grep-gate walks `src/` and fails the build if
  it leaks anywhere else.
- Always passes `--porcelain`; streams stdout+stderr through a thread-per-pipe
  tee (mirroring `src/executor/command.rs`) so output is both forwarded live to
  the presenter and fully captured for parsing.
- The three primitives became cwd-mandatory (`push_set_upstream_from`,
  `push_from`, `push_delete_from`) so native hooks fire in the correct worktree;
  the two no-cwd variants are deleted (pre-1.0 clean cut). New error contract:
  `Err` = spawn/IO failure only; a rejected push is
  `Ok(PushIo { success: false, .. })` so the composition layer classifies before
  deciding severity.
- The raw `multi_remote.rs` bypass and its `set_current_dir`/restore dance are
  gone — it routes through `push_delete_from` with an explicit cwd.
- New `pre_push_hook_exists(cwd)` resolves `core.hooksPath` (so lefthook / husky
  / pre-commit installs are detected, not just `.git/hooks`) and checks the
  executable bit.

### Pure porcelain parser (`src/git/push_porcelain.rs`)

- `parse_push_report(stdout) -> PushReport` replaces the locale-fragile
  `"Everything up-to-date"` stderr grep. A hook rejection emits zero porcelain
  ref lines — that is the gate signature. Pure function, nine unit tests against
  fixtures captured from real `git push --porcelain` runs.

### Composition (`src/core/worktree/push.rs`)

- `push_with_hooks(...)` is the single entry point. `--no-verify` ⇒ `Bypassed`;
  `manages_stage` ⇒ Path B (run the stage, then push with `verify=false` so git
  does not double-fire the hook); otherwise Path A (existence-gated synthetic
  `pre-push` phase with teed output). Collapses to a coarse
  `PushOutcome { up_to_date, hook: HookVerdict, failure }` that every caller
  branches on regardless of path.

### The `StageRunner` port (`src/core/worktree/ports.rs`)

- First port outside `src/coordinator/` (consumer-owns-port, `dyn` + sync per
  ARCHITECTURE.md). Ships with `NoopStageRunner` whose `manages_stage` returns
  `false`, so everything runs Path A today and #599 compiles and ships with zero
  #468 code. The port pins git's four-field pre-push stdin via `PushRef`.

### Reporting (CLI + TUI)

- Path A drives `CliPresenter` / `TuiPresenter` **directly** around the teed
  subprocess: one `on_phase_start("pre-push")` → one job carrying git's output →
  `on_phase_complete`. Same `JobPresenter` trait and renderers as lifecycle
  hooks — it simply drives them directly rather than through `HookExecutor`
  (a native/lefthook subprocess has no `HookContext`/`HookType`/daft job).
- The phase is **existence-gated**: a plain push with no hook shows no hollow
  header.
- TUI sync gains `DagHookPhase { Lifecycle(HookType), PrePush }` — the DAG event
  label is generalized; the closed `HookType` enum is untouched (explicit
  non-goal: no `HookType::PrePush`).

### Per-site semantics — the one deliberate behavior change

The hook always gates the push. **When a pre-push hook exists and `verify` is
on, a failed push is now an error (non-zero exit), not a warning** — a gate
saying NO must not be reduced to a warning. With no hook present (or with
`--no-verify`), the old warn-and-continue ergonomics are preserved, so the
offline UX is intact. Escalation bails are **deferred** until after worktree
setup / cd-redirect completes, so the user's shell still lands in the new
worktree even when the gated push then fails the command.

| Site | Before, on push failure | After #599 |
|---|---|---|
| `sync --push` | counted → non-zero | unchanged |
| checkout-branch auto-upstream | warn + continue | hook present ⇒ error (worktree kept + set up) |
| rename (push new name / delete old) | warn + continue | hook present ⇒ error |
| branch-delete `--remote` | recorded → non-zero | unchanged |
| multi-remote move `--push` / delete-old | warn + continue | hook present ⇒ error |

### `--no-verify` passthrough

- Added to `sync`, `checkout` (`-b` / `go` / `start`), `rename`,
  `worktree-branch` (+ `remove`), `branch-delete`, and `multi-remote move`.
  Threads to `verify = !no_verify`. Man pages + CLI reference docs regenerated.

## The up_to_date fork (resolved, not regressed)

The issue flagged a real design fork: you cannot switch to `Stdio::inherit()` to
surface hook output without emptying the captured string `sync` parses for
up-to-date, which would mislabel up-to-date branches as "Pushed successfully."
Resolved with option (b) from the issue — `--porcelain` + tee. Porcelain puts a
locale-stable verdict on **stdout** (`=\t<refspec>\t[up to date]`) while the
hook's output stays on its own streams and is teed for visibility. A second
`sync --push` after a passing push still classifies the branch as up-to-date
(covered by a dedicated scenario).

## #468 design appendix — the two-axis Path-B model (behind `StageRunner`, not built here)

The pluggable hook-runner architecture asked for in scoping lives entirely
behind the `StageRunner` port, so it costs this PR nothing:

- **Outer axis** — per repo: managed **and** trusted ⇒ Path B (daft owns the
  stage); otherwise ⇒ Path A (git dispatches). Trust gates only daft's own stage;
  it cannot gate git's native dispatch, so an untrusted repo degrades to Path A
  and the push still proceeds with the native/lefthook hook firing.
- **Inner axis** — inside Path B: a config-template plugin (a drop-in
  configuration file describing how to detect/parse/run a new runner) ⇒ a
  compiled-Rust plugin (the speed fallback when a template would be too slow) ⇒
  an opaque single-job fallback. Repo-supplied runner templates carry arbitrary
  invoke strings — the same trust surface as `daft.yml` hooks — so #468 must
  trust-gate them.

#468 supplies the real adapter; the seam already calls
`manages_stage("pre-push", repo_cwd)` and runs `run_stage(...)` before pushing
with `verify=false` (avoiding a double-fire). #600 consumes the same seam, adding
only branch→worktree resolution.

## Testing

- Unit: **1976 / 0**, clippy zero-warnings, fmt clean, man pages up-to-date.
- New unit coverage: the pure porcelain parser (9 tests), and `push_with_hooks`
  end-to-end against real temp-repo hooks with a recording presenter and a fake
  managing `StageRunner` (13 tests) — failing hook rejects and reports, passing
  hook pushes, `--no-verify` bypasses without emitting a phase, a hookless repo
  renders nothing, up-to-date survives, and Path B blocks on stage failure / does
  not re-fire on success.
- Six new YAML scenarios under `tests/manual/scenarios/push-hooks/` install a
  hook via `core.hooksPath` against the `standard-remote` fixture and assert the
  gate at each surface: `sync --push` rejected, sync passing + up-to-date
  regression, `--no-verify` bypass, checkout-branch auto-upstream rejected,
  rename remote rejected, branch-delete `--remote` rejected.
- Neighboring scenario suites re-run green (sync, branch-delete, checkout,
  rename).

## Note

Implementing the multi-remote move scenario surfaced a **pre-existing** bug:
`daft multi-remote enable` fails in the `contained` layout (it tries
`git worktree move` on the container root → "is a main working tree"). It is
unrelated to this change and deserves its own issue; site 6's gating is covered
by unit tests and shares the rename path's exact code, so no YAML scenario is
gated on that bug.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -212,7 +212,7 @@ these as `git` subcommands (e.g., `daft worktree-checkout` is
 | `daft worktree-carry <targets>`                                                                                                                                              | Transfer uncommitted changes to one or more other worktrees                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
 | `daft worktree-fetch [targets]`                                                                                                                                              | Update worktree branches from remote (supports refspec syntax source:destination)                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
 | `daft worktree-branch -m <source> <new-branch>`                                                                                                                              | Rename a branch, move its worktree, and rename the remote branch                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
-| `daft git-worktree-sync [-f] [-v\|-vv] [--rebase BRANCH [--autostash]] [--push [--force-with-lease]] [--include VALUE]... [--stat summary\|lines]`                           | Synchronize all worktrees: prune stale + update all + optional rebase + optional push (`-f`/`--prune-dirty` for dirty worktrees, `-v` hook details, `-vv` full sequential). By default, rebase and push apply only to branches you own (matching `git config user.email`). Use `--include` to add more branches: `unowned` for all, an email address for a teammate's branches, or a branch name for one specific branch.                                                                                                                     |
+| `daft git-worktree-sync [-f] [-v\|-vv] [--rebase BRANCH [--autostash]] [--push [--force-with-lease] [--no-verify]] [--include VALUE]... [--stat summary\|lines]`             | Synchronize all worktrees: prune stale + update all + optional rebase + optional push (`-f`/`--prune-dirty` for dirty worktrees, `-v` hook details, `-vv` full sequential). By default, rebase and push apply only to branches you own (matching `git config user.email`). Use `--include` to add more branches: `unowned` for all, an email address for a teammate's branches, or a branch name for one specific branch.                                                                                                                     |
 | `daft worktree-merge [SOURCE...] [--into <TARGET>] [--merge\|--squash\|--rebase\|--rebase-merge] [-s <STRAT>] [--adopt-target\|--no-adopt-target] [-y] [-r] [--set-default]` | Merge one or more source branches into a target worktree's branch. Without `--into`, the target is the current worktree. With `--into <target>`, merges into another worktree without changing directories. Multiple sources trigger octopus. `-r` removes the source worktree and branch after success. `--adopt-target` creates an ephemeral worktree when the target branch has no worktree. `--set-default` writes style and cleanup to `git config --local`. Finish with `daft worktree-merge --abort\|--continue\|--quit [<worktree>]`. |
 
 ### Adoption and Ejection
@@ -654,6 +654,18 @@ selector matching nothing warns and the run proceeds.
 (which is an inclusion filter). `--skip-hooks all` cannot be combined with
 `--trust-hooks`; a partial skip (`tag:`/`<job>`) still runs your own hooks and
 remains compatible with `--trust-hooks`.
+
+### Git pre-push Hooks on daft Pushes
+
+Separate from daft's own hooks: every daft-initiated `git push` honors the
+repo's git-level `pre-push` hook (native `.git/hooks` or `core.hooksPath`
+managers like lefthook/husky/pre-commit). The hook run is reported as a
+`pre-push` phase. A failing hook blocks the push and the command exits non-zero
+— any worktree it created or moved is still completed and usable, and the error
+names the manual recovery command. Pass `--no-verify` to the pushing command
+(`daft sync --push`, `daft start`/`go -b`, `daft rename`, `daft remove`,
+`daft multi-remote move`) to skip the hook for one invocation. `--skip-hooks`
+does NOT affect git-level hooks — it only filters daft's own jobs.
 
 ### Move Hooks
 
--- a/docs/cli/daft-multi-remote.md
+++ b/docs/cli/daft-multi-remote.md
@@ -155,6 +155,7 @@ daft multi-remote move [OPTIONS] <BRANCH>
 | `--set-upstream` | Also update the branch's upstream tracking to the new remote |  |
 | `--push` | Push the branch to the new remote |  |
 | `--delete-old` | Delete the branch from the old remote after pushing |  |
+| `--no-verify` | Skip the repo's pre-push hook on remote operations |  |
 | `--dry-run` | Preview changes without executing |  |
 | `-f, --force` | Skip confirmation |  |
 
--- a/docs/cli/git-worktree-branch-delete.md
+++ b/docs/cli/git-worktree-branch-delete.md
@@ -65,6 +65,7 @@ git worktree-branch-delete [OPTIONS] <BRANCHES>
 | `-D, --force` | Force deletion even if not fully merged |  |
 | `--local` | Only delete locally, keep remote branch |  |
 | `--remote` | Only delete the remote branch, keep local worktree and branch |  |
+| `--no-verify` | Skip the repo's pre-push hook when deleting the remote branch |  |
 | `-q, --quiet` | Operate quietly; suppress progress reporting |  |
 | `-v, --verbose` | Be verbose; show detailed progress |  |
 
--- a/docs/cli/git-worktree-branch.md
+++ b/docs/cli/git-worktree-branch.md
@@ -92,6 +92,7 @@ git worktree-branch [OPTIONS] [BRANCHES]
 | `--remote` | Only delete the remote branch, keep local worktree and branch |  |
 | `-m, --move` | Rename a branch and move its worktree |  |
 | `--no-remote` | Skip remote branch rename (only with -m) |  |
+| `--no-verify` | Skip the repo's pre-push hook on remote operations |  |
 | `--dry-run` | Preview changes without executing (only with -m) |  |
 | `-q, --quiet` | Operate quietly; suppress progress reporting |  |
 | `-v, --verbose` | Be verbose; show detailed progress |  |
--- a/docs/cli/git-worktree-checkout.md
+++ b/docs/cli/git-worktree-checkout.md
@@ -72,6 +72,7 @@ git worktree-checkout [OPTIONS] <BRANCH_NAME> [BASE_BRANCH_NAME]
 | `-s, --start` | Create a new worktree if the branch does not exist |  |
 | `-@, --at <PATH>` | Place the worktree at a specific path instead of using the layout template |  |
 | `--local` | Skip all remote operations (no fetch, no push) |  |
+| `--no-verify` | Skip the repo's pre-push hook on the automatic upstream push |  |
 | `--skip-hooks <SELECTOR>` | Skip hooks this run (all | <hook> | tag:<tag> | <job>); repeatable/comma-separated |  |
 
 ## Global Options
--- a/docs/cli/git-worktree-sync.md
+++ b/docs/cli/git-worktree-sync.md
@@ -51,6 +51,7 @@ git worktree-sync [OPTIONS]
 | `--autostash` | Automatically stash and unstash uncommitted changes before/after rebase |  |
 | `--push` | Push all branches to their remotes after syncing |  |
 | `--force-with-lease` | Use --force-with-lease when pushing (requires --push) |  |
+| `--no-verify` | Skip the repo's pre-push hook when pushing (requires --push) |  |
 | `--include <INCLUDE>` | Include additional branches in rebase/push (email, branch name, or 'unowned') |  |
 | `--stat <STAT>` | Statistics mode: summary or lines (default: from git config daft.sync.stat, or summary) |  |
 | `--columns <COLUMNS>` | Columns to display (comma-separated). Replace: branch,path,age. Modify defaults: +col,-col. Available: branch, path, size, base, changes, remote, age, annotation, owner, hash, last-commit |  |
--- a/docs/hooks/index.md
+++ b/docs/hooks/index.md
@@ -45,6 +45,16 @@ Two distinctions:
    _after_ code reaches the central repo; daft hooks run _before_. CI shifts
    left.
 
+## daft and your existing git hooks
+
+daft's hooks do not replace git's own. Every daft-initiated `git push` honors
+the repository's `pre-push` hook — whether it lives in `.git/hooks` or is
+managed by lefthook, husky, or pre-commit via `core.hooksPath`. A failing
+pre-push gate blocks the push and fails the command; its run is reported as a
+`pre-push` phase on the same surface lifecycle hooks use. Pass `--no-verify` to
+any pushing command to skip it for one invocation. See
+[Git Hooks](/reference/configuration#git-hooks) for the details.
+
 ## Where to next
 
 - **Reference:** [Lifecycle hooks](/hooks/lifecycle) — types, triggers, env
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -324,10 +324,17 @@ git config daft.hooks.worktreePostCreate.failMode abort
 
 ## Git Hooks
 
-daft's push operations are structural -- they manage branch topology as a
-side-effect of worktree management, not as user-initiated code pushes. Because
-of this, daft passes `--no-verify` on all `git push` calls, skipping any
-`pre-push` hooks configured in the repository.
+Every daft-initiated `git push` honors the repository's `pre-push` hook --
+native `.git/hooks`, or hook managers registered through `core.hooksPath`
+(lefthook, husky, pre-commit). A pre-push secret scanner or test gate that fires
+on `git push` fires on daft's pushes too. The hook run is reported as a
+`pre-push` phase with the hook's output, using the same surface as daft's
+lifecycle hooks.
+
+Pass `--no-verify` to skip the hook for one invocation. Every command that can
+push accepts it: `daft sync --push`, `daft start` / `daft go -b` /
+`git worktree-checkout -b`, `daft rename`, `daft remove` /
+`git worktree-branch -d/-D`, and `daft multi-remote move`.
 
 Remote operations are disabled by default. When enabled (via
 `daft config remote-sync --on` or by setting the individual keys), this affects:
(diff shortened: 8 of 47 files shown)


avihut:
Thanks for the review — thorough and on-point. Addressed in the latest commit; summary of what changed and what I consciously left:

### Medium — `Rejected` false-positive on transport failures → **fixed the messaging, kept the classification**

You're right that a transport failure (unreachable remote, auth) with a hook present produces zero porcelain ref lines and lands in `Rejected`. Two halves:

- **The escalation to non-zero is intended, not a bug.** #599's rule is "a failed push with a pre-push hook honored is an error, not a warning" — the offline warn-and-continue ergonomics are preserved only for hook-less repos (or `--no-verify`). So a failed push with a hook present surfacing non-zero is by design, transport or not. (You noted this is "acceptable behavior.")
- **The wrong part was the diagnostic**, which asserted "with the repo's pre-push hook in effect" — blaming the hook for what might be a network error. Fixed: `Rejected` now reads *"the repo's pre-push hook may have blocked it, or the remote was unreachable"* — naming both, asserting neither. The real git error still travels in the message.

I **declined the stderr-pattern mitigation** (`fatal: unable to access`, `ssh:`): distinguishing the causes requires parsing locale-dependent stderr, which #599 deliberately avoids ("no fragile stderr-sniffing to classify failure causes"). daft genuinely cannot tell hook-refusal from pre-negotiation transport failure on Path A (git's hook dispatch is opaque), so the honest fix is to report what it *observed* — the push died before negotiating a ref — and name the candidates, rather than guess a cause from prose. Reliable per-hook attribution is Path B (#468) territory.

### Minor — misleading message when the hook **passed** but the push failed → **fixed**

The verdict now drives the wording. `Passed` reads *"the pre-push hook passed but the remote rejected the push"* and **drops the `--no-verify` hint** (bypassing a hook that already passed wouldn't change a non-fast-forward). Threaded the verdict through `rename`'s `PushFailure::Gated` so its two sites distinguish the cases too; same treatment at checkout-branch, multi-remote (×2), and sync.

### Minor — `on_job_failure` fires for push-layer failures → **comment added**

Added a comment at that line acknowledging the coarse `pre-push: ✗` attribution is the accepted Path A tradeoff (opaque subprocess), and that the caller-facing message disambiguates via `HookVerdict`. No behavior change, as you suggested.

### Minor — `pull_in`/`rebase_in` bypass `git_command_at` → **out of scope, follow-up**

Agreed and pre-existing (not introduced here). The new push seam uses `git_command_at` correctly; hardening the older `pull_in`/`rebase_in`/`fetch` methods is a separate change I'll track as a follow-up rather than widen this PR's surface.

### Observation — grep-gate path matching → **left as-is**

The `ends_with("git")` match is on a currently-passing test, and the `assert_eq!` catches a miscount if the file ever moves. I'd rather not refactor a green invariant test for marginal robustness; noting it for if/when `remote.rs` actually relocates.

### Tests

Added three unit tests pinning the new wording (`failure_cause` doesn't blame the hook for `Rejected`, attributes `Passed` to the remote; `--no-verify` hint offered only when the hook is a candidate), and updated the three push-hooks YAML scenarios that asserted the old phrase in lockstep. Local gates green: unit **1979/0**, clippy zero-warnings, fmt, man up-to-date, and `test:manual push-hooks` **6/6**.

PR #654, opened 2026-07-01
feat(hooks): contextual per-command notices when untrusted repos skip hooks, with trust + replay suggestions

### LINKED ISSUE #596: feat(hooks): contextual per-command warnings when untrusted daft config is skipped, with trust + retroactive-replay suggestions
## Problem

When a repository has daft configuration (a `daft.yml` and/or `.daft/hooks/`) but is **untrusted**, daft loads the config, finds the hook, sees `TrustLevel::Deny`, and skips it — emitting only a `--verbose`-gated `output.debug` (`src/hooks/executor.rs:332-337` YAML path, `:437-443` legacy path; `src/output/cli.rs:119-127`). Nothing tells the user their config was found and ignored *because* the repo isn't trusted.

The one existing notice (`check_hooks_notice`, `src/hooks/executor.rs:569-590`) is incomplete on two axes:

- **It never looks at `daft.yml`** — it only stats `.daft/hooks/` files via `list_hooks`/`hook_exists` (`src/hooks/mod.rs:496-518`), so a `daft.yml`-only repo hits the `hooks.is_empty()` early-return and stays **completely silent**.
- **It's wired per-caller** — only `checkout`, `clone`, `flow_adopt` call it. `merge` (`merge.rs:1025/1104`), `sync`/`sync_shared` (`sync_shared.rs:494`), and the multi-branch clone loop (`clone.rs:628/696`) invoke `HookExecutor::execute` but call **no** notice — silent for every config type. (Same per-caller fragility class CLAUDE.md flags re the `repo remove` cd-redirect bug.)

## Desired behavior

Every command that *would have run* a lifecycle hook but skipped it on trust grounds must surface a **contextual** warning (on **stderr**, so it never pollutes `DAFT_CD_FILE`/`eval`) that:

1. **Names the specific hooks it skipped, in this command's terms.** A worktree-create says its `worktree-post-create` hooks didn't run; a remove says its `worktree-pre/post-remove` hooks didn't run; clone says `post-clone`; etc. Not a generic "hooks exist somewhere."
2. **Covers `daft.yml` as well as `.daft/hooks/`** — both config shapes, all commands.
3. **Branches on state:**
   - **Untrusted now** → "These hooks were not run because this repository isn't trusted." → suggest `git daft hooks trust`.
   - **Trusted, but the hooks were already skipped on an earlier untrusted run** → suggest running them **retroactively** so the side effects (env setup, symlinks, etc.) actually happen — e.g. "You created this worktree before trusting it; run its post-create hooks now: `git daft hooks run worktree-post-create`."

### Concrete UX (illustrative copy)

New worktree, untrusted:
```
warning: daft.yml defines a worktree-post-create hook that was NOT run — this repository isn't trusted.
         Trust it to enable hooks:        git daft hooks trust
         (then replay this worktree's setup: git daft hooks run worktree-post-create)
```
Just trusted, post-create was skipped earlier:
```
note: this repository is now trusted, but its worktree-post-create hooks were skipped when this worktree was created.
      Run them now:  git daft hooks run worktree-post-create
```

## Command → lifecycle-hook coverage

These are the commands that invoke `HookExecutor::execute` and therefore need a contextual trust notice. Today only the first group warns (and only for `.daft/hooks/`):

| Command(s) | Lifecycle hook(s) | Notice today |
|---|---|---|
| checkout / branch (create worktree) | `worktree-pre-create`, `worktree-post-create` | partial (`.daft/hooks/` only) |
| clone | `post-clone` (+ per-branch `worktree-*-create`) | partial; multi-branch loop silent (`clone.rs:628/696`) |
| flow-adopt | create hooks | partial |
| **remove / branch-delete** | `worktree-pre-remove`, `worktree-post-remove` | **none** |
| **prune** | `worktree-pre-remove`, `worktree-post-remove` (per pruned worktree) | **none** |
| **merge** | its executed hooks (`merge.rs:1025/1104`) | **none** |
| **sync / sync-shared** | its executed hooks (`sync_shared.rs:494`) | **none** |

(Verify the exact hook set each of merge/sync fires while implementing — the point is each gets a notice scoped to *its* hooks.)

## Implementation notes

- **Centralize, don't extend the per-caller notice.** Emit the warning from the trust-`Deny` arms in `HookExecutor::execute` (`executor.rs` ~332 / ~437) — every command flows through `execute`, so this closes the gap for all commands AND both config shapes in one place, and kills the "caller forgot the case" class. Pass the command/event context into the executor so the message can be hook-specific.
- **Name what was skipped.** Surface the skipped hook list (type + source: `daft.yml` vs `.daft/hooks/<file>`) on the `HookResult`/skip path so the message is specific, not generic.
- **Retroactive replay** builds on the existing `git daft hooks run` (which already bypasses trust via `with_bypass_trust(true)`, `hooks/run_cmd.rs:223`) and `git daft hooks trust` (`commands/hooks/trust.rs`). To suggest replay intelligently, record **skipped** invocations (the coordinator already logs hook invocations to `src/store/`; mark trust-skipped ones) so daft can tell "this worktree's post-create never ran" and offer it. A first cut can simply always offer the replay command after a trust action; the store-backed version makes it precise.
- **Don't double-warn** — one concise notice per command invocation, deduped across the hooks it skipped.

## Relationships

- Headline instance of the silent-failure class that **#590** (silent-failure & error-handling sweep) is chartered to find — this is its fix.
- Same "tell the user what's happening / offer the next action" philosophy as **#517** (activate auto-prompt).
- Related to the trust-wipe bug (layout-transform orphans the trust entry → `__prune-trust` deletes it): this warning would at least make that silent untrust *visible* at the next command. (Tracked separately.)

## Acceptance criteria

- [ ] Untrusted repo with a **`daft.yml`** hook → a contextual warning naming the skipped hook(s) on stderr, in every command that would have run them (create, remove, prune, clone, merge, sync)
- [ ] Untrusted repo with **`.daft/hooks/`** → same, including the merge/sync/prune/remove paths that are silent today
- [ ] Warning names the command-relevant hook(s) and suggests `git daft hooks trust`
- [ ] After trusting, the relevant command suggests retroactively running the previously-skipped hooks
- [ ] No regression to the cd-redirect / stdout contract (warnings on stderr only)
- [ ] Regression coverage: a YAML scenario (or unit test) per config shape × at least the create + remove + sync paths

avihut:
## Summary

When a repo defines hooks but isn't trusted, daft used to skip them either silently (`daft.yml` configs) or with inconsistent per-command warnings. This branch makes every trust-gated skip visible, calm, and actionable:

```
Untrusted repo — 2 daft.yml hooks not run: worktree-pre-create, worktree-post-create
   To run them, trust this repo:       git daft hooks trust
   Then replay this worktree's setup:  git daft hooks run worktree-post-create
```

### What's in here

- **One notice per command invocation**, centralized at the executor's trust gates so every command (checkout, clone, merge, sync, prune, remove) and both config shapes (`daft.yml`, `.daft/hooks/`) are covered with no per-caller wiring. TUI commands defer the notice through a process-global registry keyed by git dir and flush it to the real stderr after the TUI exits.
- **A notice, not a warning.** Emitted via a new `Output::notice` channel — stderr, no severity tag, shown even under `--quiet` — because an untrusted repo declining to run its hooks is the security model working as designed, not a problem. The line reads general → specific (trust state, count, hook names) and is styled by role on color terminals: bold state, hook names in the accent orange the hook progress UI already uses for them, suggested commands in cyan. Styling is applied after padding, so stripping the codes recovers the plain rendering byte-for-byte.
- **Skips are recorded** (`invocations` table, migration 002) so a later `git daft hooks trust` suggests replaying exactly the setup hooks that never ran, scoped per worktree. Replaying a hook clears its record; only idempotent setup hooks (`post-clone`, `worktree-post-create`) are ever suggested.
- **Executable-name parity.** Suggested commands render the way daft was invoked — `daft hooks trust` for direct invocations, `git daft hooks trust` through git — via new `cli_label()`/`daft_cmd()` helpers swept across all runtime hints (hooks notice, prompt-unavailable/migrate/re-trust warnings, `hooks status`/`trust`/`run`, multi-remote, doctor). Static clap help and man pages keep the canonical git form; a CLAUDE.md guideline prevents regressions.
- Docs: trust-and-security page (never-silent section, replay section), troubleshooting entry, and SKILL.md updated.

### Verification

- 2027 unit tests pass; clippy clean with zero warnings.
- Full manual YAML suite: **647 scenarios / 2540 steps, 0 failures**, including 10 scenarios pinning this feature: per-command notice coverage, the stderr-only channel contract, exactly-once emission across multi-branch clone, `--skip-hooks` suppression (with no replay records), replay suggestion precision and clearing, and invocation-style parity for suggestions.

Fixes #596

### DIFF
--- a/.dep-age-allowlist
+++ b/.dep-age-allowlist
@@ -13,3 +13,9 @@
 # Example:
 #   # CVE-2026-XXXX patched in 0.2.4, no older fix available
 #   cargo:somecrate@0.2.4
+
+# why: soundness advisory in anyhow <=1.0.102 (dtolnay/anyhow#451) turns the
+# cargo-deny advisories gate red; 1.0.103 is the designated fix and is 6 days
+# old at time of bump — the same security exception Dependabot PRs get
+# automatically. Safe to remove after 2026-07-03.
+cargo:anyhow@1.0.103
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -181,6 +181,19 @@ invoked, then dispatches to the matching module in `src/commands/`. Symlinks
 like `git-worktree-clone → daft` enable Git subcommand discovery. Shortcut
 aliases (e.g., `gwtco`) are resolved in `src/shortcuts.rs` before routing.
 
+**Referring to the daft executable in output**: Runtime hints and suggested
+commands must render the executable the way the user invoked it —
+`daft hooks trust` for direct invocations (`daft`, `daft-go`, …),
+`git daft hooks trust` for git-style ones (`git daft`, `git worktree-checkout`,
+…). Use `daft::cli_label()` / `daft::daft_cmd(...)` (src/lib.rs); never hardcode
+`git daft` into a runtime string. Static text keeps the canonical `git daft`
+form: clap `about`/`long_about` (man pages are pre-generated from them) and
+`docs/` pages; SKILL.md uses the direct `daft` form (agents invoke the binary
+directly). Under `cfg!(test)` `cli_label()` always returns the canonical
+`git daft` — unit tests share one process, so argv-derived labels would flip
+assertions order-dependently; test the classification itself through
+`label_for_argv0`.
+
 **Shell integration**: `daft shell-init` generates shell wrappers that create a
 temp file and pass its path via `DAFT_CD_FILE`. When set, commands write the cd
 target to that file, and the wrapper reads it after the command finishes to `cd`
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,9 +103,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "approx"
--- a/SKILL.md
+++ b/SKILL.md
@@ -605,14 +605,17 @@ daft hooks dump         # Show fully merged configuration
 daft hooks run <type>   # Manually run a hook (bypasses trust)
 ```
 
-When a command skips hooks because the repo is untrusted, it prints one stderr
-warning naming the skipped hooks and suggesting `daft hooks trust` (suppressed
-by an explicit `--skip-hooks`). Each skip is recorded; a later
-`daft hooks trust` lists precise replay commands (`daft hooks run post-clone` /
-`daft hooks run worktree-post-create`) for the worktrees whose setup hooks never
-ran — run them inside each listed worktree. If an agent sees that warning,
-trusting and replaying is the way to get the worktree into its fully set-up
-state.
+When a command skips hooks because the repo is untrusted, it prints one plain
+stderr notice — e.g.
+`2 daft.yml hooks not run: worktree-pre-create, worktree-post-create — this repo isn't trusted.`
+— naming the skipped hooks and suggesting `daft hooks trust` (suppressed by an
+explicit `--skip-hooks`). It is an untagged notice, not a `warning:`, because an
+untrusted repo declining to run hooks is by design. Each skip is recorded; a
+later `daft hooks trust` lists precise replay commands
+(`daft hooks run post-clone` / `daft hooks run worktree-post-create`) for the
+worktrees whose setup hooks never ran — run them inside each listed worktree. If
+an agent sees that notice, trusting and replaying is the way to get the worktree
+into its fully set-up state.
 
 ### Manual Hook Execution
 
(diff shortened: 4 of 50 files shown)


avihut:
Review triage — all three findings dispositioned:

**Fixed in aa00ea3 — delete+insert atomicity (Medium).** `record_skipped_invocation` now wraps the pair in the store's existing `with_write_txn` helper (which also rolls back on error, avoiding the open-transaction-on-pooled-connection hazard a raw `BEGIN`/`COMMIT` would add). The stale comment claiming single-writer-pool safety is gone. Regression test: `record_skipped_invocation_is_atomic_across_pools` runs two stores (two pools — the two-process shape) through 200 barrier-synced one-shot rounds. A single long hammer only samples the race once — only the last pairs on a key decide the end state — so the test uses many independent rounds instead; it fails 3/3 against the unwrapped implementation and passes deterministically with the transaction.

**Addressed as a comment in 296864a — `merge` replay fallback (Minor).** The `a.max(b)` arm now states its reachable domain: `replay` is only ever populated with the two idempotent setup hooks, so both sides there can only be `post-clone`; `max()` stays for determinism if the set grows.

**Declined — static sentinel invocation ID (FYI).** Beyond the schema change the review itself flagged, the suggestion is unsound as stated: the PK is `(repo_hash, invocation_id)`, so a fixed sentinel would make skip rows for *different* hook types/worktrees in the same repo collide, and the upsert would clobber unrelated skip records. The per-skip UUID mint stays.

Issue #669, opened 2026-07-05
builds at a tagged release commit drop daft_dev_build, disabling DAFT_*_DIR test isolation — pre-push fails and tests write the real state dir every post-release window


avihut:
## Summary

One incident, one root cause, three fixes tracked as a checklist (convert items 3–4 to sub-issues if they outgrow a PR each).

After v1.18.0 shipped (2026-07-05), every fresh build in a worktree whose HEAD sits exactly on the tagged release commit silently compiles in **release mode**: `build.rs`'s `is_release` tag heuristic suppresses `cargo:rustc-cfg=daft_dev_build`, and `daft_state_dir()` / `daft_config_dir()` / `daft_data_dir()` **compile out** the `DAFT_*_DIR` override branch (`src/lib.rs`). Ten unit tests assert override behavior → `mise run test:unit` fails → the pre-push hook rejects every push from that worktree. Worse, with overrides ignored, all test tiers (unit, rust integration, YAML sandbox) silently write the **real** `~/.local/state/daft`.

## Root cause (verified)

- `build.rs`: `is_release` = any tag at HEAD matches `CARGO_PKG_VERSION` — and it gates **both** the version-display string and the `daft_dev_build` cfg ("A build is 'dev' when it's not a release AND has a git repo").
- `src/lib.rs`: the env-override branch exists only under `cfg!(daft_dev_build)`.
- The failing assertions show the mechanism directly: `left: "/Users/avihu/.local/state/daft"`, `right: "/tmp/test-daft-state"` — `set_var` works; the reading code path isn't in the binary.

Verify in seconds:

```
git tag --points-at HEAD          # a vX.Y.Z matching Cargo.toml → this bug
grep -c 'rustc-cfg=daft_dev_build' target/debug/build/daft-*/output   # 0 in the freshest output = release-mode build
```

Observed timeline (v1.18.0): tag created upstream 20:09:53, auto-followed into the local repo 20:10:31 mid-`daft sync`. master's pre-push built ~1 minute **before** the tag landed → dev mode → green → pushed. daft-357 rebased **after** → release mode → 10 deterministic failures, reproduced across two clean rebuilds and an independent non-TTY rerun. Historical build-script outputs in `master/target` all show `daft_dev_build` emitted with `(dev master <sha>)` display; the post-tag outputs show not-emitted with a clean `1.18.0`.

[...]

PR #670, opened 2026-07-05
fix: decouple daft_dev_build from release-tag proximity

avihut:
## Summary

Fixes the root cause of #669 (checklist items 1–2): a build at a HEAD carrying the matching version tag — i.e. **every post-release window** — silently compiled in release mode, dropping the `daft_dev_build` cfg. That compiled out the `DAFT_{CONFIG,DATA,STATE}_DIR` override branch in `src/lib.rs`, deterministically failing ten override unit tests (so the pre-push hook rejected every push from a worktree on the release commit) and pointing *all* test tiers — unit, integration, YAML sandbox, and the checked-in `shared-env.sh` dev sandboxing — at the real `~/.local/state/daft` (~24.9k litter dirs accumulated since April).

## Changes

- **`build.rs`** — the dev cfg is now gated only on "git checkout present" plus the explicit `DAFT_BUILD_RELEASE` signal; the tag-at-HEAD heuristic remains for `DAFT_VERSION_DISPLAY` cosmetics only (renamed `is_release` → `is_tagged_release` to make the split self-documenting). `rerun-if-env-changed=DAFT_BUILD_RELEASE` already existed.
- **`src/lib.rs`** — `cfg!(daft_dev_build)` → `cfg!(any(daft_dev_build, test))` in the three dir resolvers: unit tests are structurally immune to build-script detection, whatever it evolves into.
- **`.github/workflows/release.yml`** — workflow-wide `DAFT_BUILD_RELEASE: "1"` so shipped binaries keep ignoring the env redirects (they build from a git checkout, so without this they'd gain the dev cfg — the security property the old tag check was accidentally providing). Manually maintained block with a WARNING comment, following the existing `publish-homebrew-formula` convention.
- **`.github/workflows/test.yml` + `scripts/check-release-build-env.sh`** — unconditional `release-env-guard` job asserting the env block survives `dist generate` regenerations. Unconditional because the dangerous PR (a regenerated `release.yml`) touches no path the `changes` filters watch.
- **`docs/about/contributing.md`** — documents the new dev-build definition (git checkout without `DAFT_BUILD_RELEASE`; tags play no part).

## Regression coverage

The ten existing override tests (`test_daft_{state,config,data}_dir_override*`, `*_rejects_relative_path`, `test_render_centralized_template`, `list_all_repo_hashes_filters_non_uuid_dirs`) are the reproduction — the bug is a compile-time cfg decision, so no single test binary can observe both sides. This branch's base **is** the tagged `v1.18.0` commit, which allowed verifying both sides at the real bug condition without moving HEAD (working-tree edits only):

| | before | after (HEAD still tagged) |
|---|---|---|
| `mise run test:unit` | 2017 passed, **11 failed** | **2028 passed, 0 failed** |
| `grep -c daft_dev_build target/debug/build/daft-*/output` | 0 (cfg dropped) | 1 (cfg emitted) |
| `DAFT_BUILD_RELEASE=1` build | — | cfg **not** emitted, clean `1.18.0` display |
| `scripts/check-release-build-env.sh` | — | passes; fails (exit 1) on a copy with the env stripped |

`mise run fmt` and `mise run clippy` clean.

## Notes

- Part of #669 — items 3 (state-dir GC for the accumulated litter) and 4 (reader-path CANTOPEN/BUSY UX) are follow-ups in `area:store`.
- After this merges, worktrees sitting on a tagged release commit regain a green pre-push on their next rebuild, and test-state pollution stops accumulating.


avihut:
Addressed the review in `eb70f354` + `24f85ce3`:

- **`test_daft_state_dir_empty_falls_back` added** — the coverage gap flagged above; all three dir resolvers now have symmetric empty-string fallback tests (suite: 2029 passed / 0 failed).
- **`cfg(test)` scope documented** — the three resolver docs now state that `cfg(test)` covers only this crate's unit tests, while integration tests exercise the built binary and rely on the build.rs cfg.
- Not taken: the guard's nested-`env:` edge case — as the review notes, the guard's job is absence detection and a misplaced key is recoverable at build time; left as-is.

Also in this push, unrelated to the review: `deps: plist 1.8.0 → 1.10.0`, pulling `quick-xml 0.41.0`. Two DoS-class RustSec advisories against `quick-xml <0.41` (RUSTSEC-2026-0194 / RUSTSEC-2026-0195) published after this branch's CI baseline were failing `dep-license-check` for every rust-touching PR. quick-xml only reaches daft via `plist → syntect` (bundled tmTheme parsing, not untrusted XML), so exposure was theoretical — the bump unblocks the gate. Both versions are younger than the 7-day cooldown, so they carry dated `.dep-age-allowlist` exceptions (droppable after 2026-07-11), the same path Dependabot security PRs take automatically. `cargo-deny` green locally: advisories ok, bans ok, licenses ok, sources ok.

PR #672, opened 2026-07-06
fix(sync): graceful two-stage Ctrl+C cancellation; harden alias capture against job-control stops

### LINKED ISSUE #663: pre-push unit tests can be job-control-stopped under a real TTY, wedging daft sync --push beyond ctrl+c
## Summary

A `daft sync -v --columns=+size --rebase master --autostash --push --force-with-lease` run wedged indefinitely (12+ minutes) while the `daft-357/feat/repo-catalog` worktree's pre-push hook (`lefthook` → `mise run test:unit`) was running. Ctrl+C did nothing. The unit-test process group had been **stopped by terminal job control** (`T` state) — including the in-flight `$SHELL -i` alias capture *and the daft test process hosting its 10-second watchdog* — so no timeout could ever fire and no signal short of SIGKILL could make progress. Recovery required `kill -KILL -<pgid>` on three process groups.

Field-observed on daft v1.18.0 (`02af3209`), macOS (Darwin 25.4.0), zsh login shell.

## Observed process state (live capture before recovery)

```
PID    PPID   PGID   STAT  ELAPSED  COMMAND
993    899    993    Ss+   —        /bin/zsh --login                      ← terminal; owned tty fg at capture time
65365  993    65365  T     12:57    git-worktree-sync -v --columns=+size --rebase master --autostash --push --force-with-lease
67624  65365  65365  T     12:46    git -C …/daft-357/feat/repo-catalog push --porcelain --force-with-lease origin daft-357/feat/repo-catalog
67802  67624  65365  T     12:44    /bin/sh /Users/avihu/Projects/daft/.git/hooks/pre-push origin git@github.com:avihut/daft.git
67806  67802  65365  T     12:44    lefthook run pre-push origin git@github.com:avihut/daft.git
67815  67806  67815  S     12:44    mise run test:unit                    ← own pgrp; alive, blocked in wait()
67941  67815  67941  T     12:41    bash mise-tasks/test/unit             ← own pgrp; STOPPED
67984  67941  67941  T     12:40    cargo test --lib --tests --locked -- --test-threads=4
846    67984  67941  T     10:42    target/debug/deps/daft-3d54eccf2c737f75 --test-threads=4
11674  846    67941  T     10:33    bash -i -c alias -p >"$__DAFT_ALIAS_OUT" 2>/dev/null; declare -f >"$__DAFT_FN_OUT" 2>/dev/null
```

Key structural facts:

- `mise` and its task shell get **their own process groups** (67815, 67941) — background groups with respect to the session's controlling terminal. The tty can never deliver keyboard signals to them.
- The entire cargo-test group (67941) is `T` (stopped). A stopped process acts on **only** SIGKILL and SIGCONT; SIGINT/SIGTERM merely queue as pending.
- PID 11674 is unambiguously `AliasCache::capture` (`src/core/worktree/exec/alias_cache.rs` — the `__DAFT_ALIAS_OUT` / `__DAFT_FN_OUT` env-var names and command body match `capture()` exactly). The `alias -p` / `declare -f` pair is the **bash** capture path, and the user's real `$SHELL` is zsh (`alias -L` / `typeset -f`), so this spawn came through the `capture_survives_rc_file_stdout_pollution` fixture wrapper, not a real-user capture.
- The capture child was still alive when frozen, i.e. the group-wide stop landed **within the 10 s `CAPTURE_TIMEOUT` window** after the spawn — the watchdog loop in `capture()` was itself stopped before it could fire. The safety net froze together with the thing it guards.

## Why Ctrl+C could not recover it (three independent reasons)

1. **Wrong group.** Ctrl+C delivers SIGINT to the terminal's *foreground* process group only. The wedged subtree lives in background groups (67815/67941); tty-generated signals can never reach it.
2. **Stopped processes don't act on SIGINT.** Everything reachable was already `T`; a delivered SIGINT queues as pending and is only acted on after a SIGCONT — which nothing ever sent.
3. **`wait()` chain.** sync → git push → hook → lefthook → mise → task shell → cargo → test binary: each level blocks in `wait()` on its child, so even where a signal did land, the pipeline could not unwind while the deepest group stayed stopped.

Recovery that worked: `kill -KILL -65365; kill -KILL -67815; kill -KILL -67941` (SIGKILL is the only terminating signal a stopped process acts on; targeting the *group* reaches the background members Ctrl+C cannot).

## Trigger: confirmed vs. open

**Confirmed:** the cargo-test process group received a job-control stop (SIGTTIN/SIGTTOU class) as a *group*, within seconds of the alias-capture spawn, while running under a real controlling terminal (interactive push). Unit tests only meet this hazard when invoked from a terminal — CI and non-TTY invocations are immune, which is exactly the known pattern that interactive-prompt tests hang under a real TTY but pass in non-TTY CI.

**Not yet pinned down:** which process performed the tty access that drew the signal. Candidates:

- a parallel test thread (`--test-threads=4`) driving an interactive-prompt code path that opens `/dev/tty` directly — a background-group read of the controlling terminal raises SIGTTIN for the whole group;
- the `-i` shell's own startup tty handling (less likely: `capture()` nulls all three stdio fds, which should disable bash job control).

A follow-up manual `git push` from the same worktree did **not** hang (the suite ran to completion), so the trigger is nondeterministic — consistent with thread-interleaving-dependent tty access. Pinning it needs a PTY-harness repro: run the lib suite under an owned PTY with the suite placed in a background pgrp, and bisect for which test stops the group.

## Impact

Every `git push` / `daft sync --push` from a real terminal runs `mise run test:unit` via the pre-push hook, in a background process group, under the user's controlling tty. Any tty-touching unit test can wedge any push — un-interruptibly — and under `daft sync` it presents as *daft itself* hanging mid-sync with Ctrl+C dead. Users without job-control forensics will reach for a new terminal and a hard kill, or reboot.

## Proposed fixes (layered)

1. **`AliasCache::capture`: spawn the shell in its own process group.** Add `.process_group(0)` (`std::os::unix::process::CommandExt`, stable) to the `Command` in `capture()`. No `unsafe`, so Critical Rule #4 holds. If the child then self-stops on tty access, it freezes *alone*: daft's watchdog keeps running, kills it at the 10 s deadline, and `daft exec` falls back to the rc-less path. Cheap; worth shipping regardless of the root trigger.
2. **Audit unit tests for controlling-terminal access.** No unit test may read/write `/dev/tty` or spawn `-i` shells against the invoking terminal. Interactive prompts must have interactivity injected from the caller and be verified under an owned PTY (existing project guidance for interactive-prompt tests).
3. **Optional belt-and-braces:** detach the pre-push test task from the controlling terminal (setsid-equivalent spawn for `mise run test:unit` in the hook), so a stray tty access can never stop the suite mid-push.

## Regression test

Per repo policy the fix must ship with a failing-first reproduction: a PTY-based test that runs the offending path (at minimum, `capture()` under a PTY while its process group is backgrounded) and asserts completion within the deadline instead of a `T`-state wedge.

avihut:
Fixes #663

## What

`daft sync` gains graceful two-stage cancellation, and the #663 wedge class (job-control-stopped pre-push hook subtrees that outlive Ctrl+C) becomes recoverable with a single keypress:

- **First Ctrl+C / SIGTERM / SIGHUP** — stop scheduling new work, actively tear down running git subtrees **by process group** across all phases (fetch/pull/rebase/push), auto-`rebase --abort` any mid-rebase worktree, print a partial summary, exit 130. Stopped (`T`-state) groups are unstuck with TERM-then-CONT — the CONT delivers the queued TERM.
- **Second Ctrl+C** — immediate SIGKILL cascade over every recorded and freshly discovered descendant group (SIGKILL is the one terminating signal a stopped process acts on directly), exit 130.
- Works in the TUI path (raw-mode key event), the sequential `-vv` path, and non-TTY invocations (signal-delivered).
- A push whose isolated group job-stops on an interactive auth prompt (`/dev/tty` read from a background group) is detected via stop probes and fails that worktree's push with an ssh-agent/credential-helper hint instead of wedging the run.

## Mechanics

- **`src/git/cancel.rs` (new)** — `CancelFlag` (0→1 soft→2 hard, moved from exec with a re-export), `ps`-snapshot process-tree walk, cumulative TERM+CONT / KILL cascades that re-walk every 500ms (orphans reparent to PID 1; late-spawned groups appear after the cancel begins), stop probes, and survivor-group reporting (`kill -KILL -<pgid>` hints — the incident's manual forensics as first-class UX).
- **Supervision is opt-in.** `GitCommand::with_cancel` threads the flag into the fetch/pull/rebase/push seams; with no flag the seams keep classic blocking behavior in the caller's process group, so clone/checkout/prune fetches and autopush keep their terminal credential prompts and Ctrl+C reach unchanged.
- `run_push` polls via `ChildSupervisor` and gates on pipe-drain completion — TERM-immune hook descendants hold git's pipe write-ends past git's death, and returning early would leave nobody watching the flag (the original wedge shape).
- DAG executor sweeps Pending→Cancelled with per-task completion events; the TUI renders cancelled runs; `ctrlc` gains the `termination` feature (SIGTERM/SIGHUP parity) and `prompt.rs` exits 130 instead of 0 when signalled.
- **`AliasCache::capture` hardening** — the capture shell now runs in a fully detached session via the `daft __capture-aliases` spawn-self trampoline (safe `nix` `setsid()`, then `exec` the shell in place). A hostile rc that stops its own group can never freeze daft's watchdog, and — the subtle half — an interactive shell whose session still holds daft's tty never job-stops *itself* fighting for foreground (`bash -i` force-opens `/dev/tty` and does the SIGTTIN dance even with job control disabled; a session with no controlling terminal skips it entirely).

## Tests

- Unit: 15 cancel-module tests (stopped-group unsticking, TERM-immune KILL escalation, orphaned pipe holder, group-placement pins for supervised/unsupervised spawns) plus capture regressions (self-stopping shell; rc stdout pollution, which doubles as the tty-dance regression when the suite runs from a real terminal).
- Integration: `tests/integration/test_sync_cancel.sh` — graceful single-SIGINT teardown, TERM-immune second-SIGINT force-kill, the #663 incident repro (a verified `T`-state foreign-group descendant unstuck by one SIGINT), SIGTERM parity. A pty-wrapped test in `test_worktree_exec.sh` covers the capture trampoline's production dispatch (`script(1)` fabricates the controlling terminal CI lacks); verified failing-first.
- Manual walk: `test-plans/sync-cancellation.md`.


avihut:
Review findings addressed in 27e1c569:

- **ctrlc handler conflict (bug)** — consolidated: `prompt.rs` arms a `PROMPT_ACTIVE` takeover state (RAII guard) around the interactive read, and sync's process-global handler calls `prompt::exit_if_prompt_active()` before escalating, honoring the prompt's exact cancel contract (message + exit 130). A SIGTERM mid-prune-prompt now exits immediately instead of hanging until a keypress.
- **`git_command_at` violation (bug)** — `execute_local_branch_update`'s `merge-base`/`branch -f` now route through `git_command_at(git_dir)` so an inherited `GIT_DIR` can't retarget them when sync runs inside a git hook.
- **`ps` probe cost (minor)** — documented on `STOP_PROBE_EVERY`: ≤2N short-lived `ps` forks/sec while children run un-cancelled, with the shared-snapshot lever noted if N ever grows painful.
- **`tty_stopped` race (minor)** — documented on `STOP_STREAK`: why two probes at 500 ms, and why the residual stop-resume-exit race is accepted.

Gates re-verified: fmt/clippy clean, 2045 unit tests (new guard regression test included), sync + sync-cancel integration suites green.

PR #673, opened 2026-07-06
fix: detect multi-commit squash merges in the gone-branch merge verifier

### LINKED ISSUE #662: prune/sync misdetect squash-merged multi-commit branches as gone-but-unmerged, requiring --force
## Symptom

Since v1.17.0, `daft prune` and `daft sync` no longer remove worktrees/branches of feature branches that were **squash-merged** to the default branch (the normal PR flow for this repo), unless `--force` is passed:

```
warning: Skipping feat/multi: remote branch is gone but the local branch is not merged into master (use --force to delete anyway)
warning: Skipped 1 gone-but-unmerged branch (use --force to delete): feat/multi
```

The branch is genuinely merged — its full content is on `master` — so routine post-merge cleanup now requires `--force`, which defeats the point of the guard (users get trained to always pass `--force`, re-exposing them to the #628 data-loss scenario the guard exists to prevent).

## Root cause

PR #629 added the (intentional, correct) gone-but-unmerged guard: every gone branch is verified with `is_branch_merged` before deletion (`src/core/worktree/prune.rs`, the `!params.force` block in `prune_single_branch`).

The verifier (`src/core/worktree/merged.rs::is_branch_merged_into`) does:

1. `git merge-base --is-ancestor` — detects regular/FF merges.
2. `git cherry <target> <branch>` — detects squash merges; every line must start with `-`.

`git cherry` matches commits by **per-commit patch-id**. A branch with N > 1 commits squashed into a single commit on the target has no individual commit whose patch-id matches the combined squash commit, so every line comes back `+` and the branch is classified unmerged. Single-commit branches whose diff was altered during merge (conflict resolution, maintainer edits) misfire the same way.

Since PRs here are always squash-merged and real feature branches almost always have more than one commit, effectively every merged branch now survives prune/sync.

## Reproduction

In a throwaway repo with a local bare remote:

1. `feat/single`: one commit, squash-merged to master, remote branch deleted.
2. `feat/multi`: two commits, squash-merged to master, remote branch deleted.

```
$ git cherry master feat/single
- c2925e4...          # patch-id matches the squash commit -> detected merged
$ git cherry master feat/multi
+ a74e2be...
+ da369b4...          # no per-commit patch-id match -> misdetected unmerged

$ daft prune
warning: Skipping feat/multi: remote branch is gone but the local branch is not merged into master (use --force to delete anyway)
 * — pruned feat/single — removed worktree, local branch, remote tracking branch
```

`daft sync` skips identically (shared `prune_single_branch` path). `daft prune --force` removes it.

## Why tests missed it

The #629 scenarios cover a genuinely-unmerged branch being kept (`prune/prune-skips-gone-but-unmerged.yml`) and a branch with *no unique commits* being removed (`prune/prune-merged-stale-pristine-removes-silently.yml`). There is no scenario where a branch with real commits is actually squash-merged before its remote is deleted — the common real-world case.

## Candidate fix directions

The guard should stay; the squash detector needs to recognize "the branch's cumulative diff landed as one commit":

- Compare the patch-id of the branch's **combined** diff (`git diff <merge-base>..<branch>` piped to `git patch-id --stable`) against patch-ids of recent default-branch commits, instead of / in addition to per-commit `git cherry`.
- Or a tree-level subsumption check: `git merge-tree --write-tree <default> <branch>` yielding the default branch's current tree means the branch contributes nothing new (requires git >= 2.38; also catches conflict-edited merges).

Regression test required either way: a YAML scenario with a multi-commit branch squash-merged to the default branch, remote deleted, `daft prune` (no `--force`) must remove the worktree and branch.

## Origin

- Guard introduced by #629 (fixes #628), first released in v1.17.0.
- `src/core/worktree/merged.rs` — `is_branch_merged_into`
- `src/core/worktree/prune.rs` — unmerged skip in `prune_single_branch`
- `src/git/refs.rs` — `cherry()` wrapper

avihut:
## Problem

PR #629 added a data-loss guard: `prune`/`sync` verify a gone branch is actually merged before deleting its worktree, and `daft remove` applies the same check (Check 4). The shared verifier ran two checks — `merge-base --is-ancestor` (regular/FF merges) and `git cherry` (squash merges via per-commit patch-id equivalence).

`git cherry` matches **per-commit** patch-ids, so a branch of N>1 commits squashed into a single target commit never matches: every merged multi-commit branch was misclassified as unmerged. Since PRs here are always squash-merged, routine post-merge cleanup demanded `--force` — which trains users to always force, re-exposing the very #628 scenario the guard exists to prevent.

## Fix

Add a third detection step to `is_branch_merged_into`, reached only when the two cheaper checks say "not merged" — a cumulative-diff squash probe (the `git-delete-squashed` technique):

1. `git merge-base <target> <branch>` → fork point (no merge base ⇒ unrelated histories ⇒ not merged)
2. `git commit-tree <branch>^{tree} -p <base>` → synthetic unreferenced commit whose diff ≡ the branch's total diff since forking
3. `git cherry <target> <synthetic>` → a `-` line means a patch-equivalent commit exists on the target ⇒ squash-merged

The probe stays conservative: "merged" is only ever reported when an actual patch-equivalent commit exists on the target, so the guard loses nothing. All three consumers (prune, sync, branch-delete) benefit through the shared verifier. The probe's dangling commit object is swept by `git gc`; identity for `commit-tree` is env-injected so it works without configured `user.name`/`user.email`.

Known remaining gap (documented in code docs): a squash commit whose diff was altered during merge (conflict resolution, maintainer edits) still reads as unmerged — deliberately conservative; a `git merge-tree` subsumption check is a possible follow-up.

## Tests

- 5 unit tests in `src/core/worktree/merged.rs` — the regression case (multi-commit squash → must detect merged; fails without the fix), single-commit squash, regular merge, genuinely-unmerged stays unmerged, unrelated histories.
- 2 YAML scenarios: `prune/prune-removes-squash-merged-multi-commit.yml` and `branch-delete/deletes-squash-merged-multi-commit.yml` — end-to-end cleanup without `--force`.
- Existing guard scenarios (`prune-skips-gone-but-unmerged`, `refuses-unmerged`, `force-unmerged`) stay green.

Fixes #662

### DIFF
--- a/src/core/worktree/merged.rs
+++ b/src/core/worktree/merged.rs
@@ -128,15 +128,14 @@ mod tests {
     use super::*;
     use serial_test::serial;
     use std::path::{Path, PathBuf};
-    use std::process::{Command as ShellCommand, Stdio};
+    use std::process::Stdio;
 
     /// Test-only helper: run `git` quietly in `path` and panic if it fails.
+    /// Routed through `git_command_at` so all inherited `GIT_*` vars are
+    /// cleared (per the project's test-hygiene rule), not just GIT_DIR.
     fn git_ok(path: &Path, args: &[&str]) {
-        let status = ShellCommand::new("git")
+        let status = crate::utils::git_command_at(path)
             .args(args)
-            .current_dir(path)
-            .env_remove("GIT_DIR")
-            .env_remove("GIT_WORK_TREE")
             .stdout(Stdio::null())
             .stderr(Stdio::null())
             .status()
--- a/src/core/worktree/merged.rs
+++ b/src/core/worktree/merged.rs
@@ -39,8 +39,21 @@ pub fn is_branch_merged(
 
 /// Check whether `branch` has been merged into `target`.
 ///
-/// Two-step: `merge-base --is-ancestor` detects regular merges; `git cherry`
-/// detects squash merges (all lines start with `-`).
+/// Three checks, cheapest first:
+///
+/// 1. `merge-base --is-ancestor` — regular and fast-forward merges.
+/// 2. `git cherry` — per-commit patch-id equivalence (all lines `-`).
+///    Catches rebase merges and single-commit squashes, but NOT a
+///    multi-commit branch squashed into one target commit: no individual
+///    commit's patch-id matches the combined squash commit (#662).
+/// 3. Cumulative-diff squash probe — a synthetic commit of the branch's
+///    whole tree parented at the merge base represents the branch's total
+///    diff as one commit; `git cherry` on that finds the squash commit.
+///
+/// Known remaining gap: a squash commit whose diff was altered during the
+/// merge (conflict resolution, maintainer edits) matches no probe and is
+/// still reported unmerged — deliberate, since every check here must be
+/// conservative: reporting "merged" is what authorizes deleting the branch.
 pub fn is_branch_merged_into(git: &GitCommand, branch: &str, target: &str) -> Result<bool> {
     // Step 1: Check if branch is an ancestor of the target (regular merge)
     let is_ancestor = git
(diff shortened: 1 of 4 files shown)


avihut:
Review addressed in 7ed63754:

- **Item 1 (probe dates)** — fixed. `commit_tree` now injects fixed epoch `GIT_AUTHOR_DATE`/`GIT_COMMITTER_DATE`, so repeated probes of the same tree+parent dedupe to one SHA instead of accumulating a dangling object per run.
- **Item 2 (test env scrub)** — fixed. The `git_ok` test helper now routes through `utils::git_command_at`, clearing the full `GIT_*` set per the test-hygiene rule rather than hand-clearing two vars.
- **Item 3 (gitoxide branches)** — no change, as noted in the review: consistent with the existing `rev_parse`/`merge_base_is_ancestor` pattern in `refs.rs`; a future gitoxide sprint picks these up together.
- **Double merge-base call** — leaving as-is per the review's own assessment: cold path (both cheaper checks must fail first), one subprocess saved isn't worth widening the step-1 helper's contract.

Issue #668, opened 2026-07-05
Worktree size computation is slow in large projects — optimize the walk (follow-up to #665)


avihut:
## Context

Follow-up to #665. Once that fix lands, the worktree size walk only runs when the user opts in (`--columns +size`, `--sort -size`, or an activity/mtime-driven sort) — but when it does run, it is still slow: computing sizes is a full recursive walk of every worktree, and it dominates the command's wall time.

Measured while diagnosing #665 (daft repo, 5 worktrees, ~40 GB combined, Apple Silicon / APFS): the walk costs ~1.6 s with a warm FS cache and ~4 s cold, and it is a pure `lstat`/`readdir` syscall storm (`sys ≫ user`). In `node_modules`-heavy projects the same walk is in the 30 s class. `daft list --columns +size` will keep paying this until the walk itself gets faster.

## Current implementation

- `compute_directory_size` is a **single-threaded** recursive `read_dir` + `symlink_metadata` walk per worktree, with hard-link dedup via a `(dev, ino)` set — [`src/core/worktree/list.rs#L688-L718`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list.rs#L688-L718).
- The streaming collector spawns **one unbounded thread per worktree** ([`src/core/worktree/list_stream.rs#L112-L121`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list_stream.rs#L112-L121)), so N worktrees mean N simultaneous full-tree walks contending for the same disk.
- Results are recomputed from scratch on every invocation. The existing `cell_cache` cannot help: it is content-addressed by git SHAs, and the working tree (the thing being sized) has no content address.

## Candidate optimizations

Roughly in order of effort:

1. **Bound cross-worktree concurrency.** Replace thread-per-worktree with a small worker pool for the SIZE cluster (the cheap git clusters can stay wide). Avoids N-way disk thrash; helps most when many worktrees exist. Low risk, no new deps.
[...]

PR #675, opened 2026-07-06
fix(list): collect only the fields the live view renders or sorts by

### LINKED ISSUE #665: daft list takes seconds to exit after the table is fully rendered (live path walks every worktree for hidden Size/Mtime)
## Symptom

`daft list` renders the complete table almost instantly, then the process sits there for several seconds before exiting. In the daft repo itself the whole command takes ~4 s wall; in larger projects it has been observed taking up to **30 s**. The table content is fully populated long before the prompt returns — the tail is pure dead time.

```
❯ daft list
     Branch                                 Path                        ...  Commit
     daft-357/feat/repo-catalog             ...                              25m chore: release v1.18.0 (#660)
     ...
  ✦  master                                 master                          25m chore: release v1.18.0 (#660)
    ~/projects/daft ─────────────────────────────────────── 4s   20:35:06
```

Environment: daft 1.18.0, macOS 26.4.1 (ARM64), local filesystem.

## Root cause

The live (TTY) list path asks the streaming collector for **every** field — including `SIZE` and `MTIME`, which no default column displays — and the renderer cannot exit until **all** collector workers finish. So after the visible cells land, the process spends the rest of its life recursively walking the full directory tree of every worktree to compute sizes nobody asked for, then throws the result away.

The chain, at v1.18.0 (`02af3209`):

1. On a TTY without `--format`/`DAFT_NO_LIVE`, `list::run` dispatches to `run_live` — [`src/commands/list.rs#L187-L191`](https://github.com/avihut/daft/blob/02af3209/src/commands/list.rs#L187-L191).
2. `run_live` requests `fields: FieldSet::ALL` from the collector — [`src/commands/list_live.rs#L170-L179`](https://github.com/avihut/daft/blob/02af3209/src/commands/list_live.rs#L170-L179). `ALL` includes `SIZE` and `MTIME` ([`src/core/worktree/info_field.rs#L43-L44`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/info_field.rs#L43-L44)).
3. Each per-worktree worker therefore runs the SIZE cluster — annotated *"slowest cluster"* in the source — which is `compute_directory_size`: an unfiltered recursive walk (`read_dir` + `symlink_metadata` on every entry, `target/`, `node_modules/`, everything) — [`src/core/worktree/list_stream.rs#L322-L327`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list_stream.rs#L322-L327), [`src/core/worktree/list.rs#L688-L718`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list.rs#L688-L718). The MTIME cluster then re-runs `count_changed_files` per worktree ([`list_stream.rs#L329-L340`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list_stream.rs#L329-L340)).
4. The renderer's exit condition for phase-less commands is `live.collection_complete` ([`src/output/tui/state.rs#L205-L207`](https://github.com/avihut/daft/blob/02af3209/src/output/tui/state.rs#L205-L207)), which is only set by `WorktreeInfoCollectionDone`, which `CollectorHandle::join` emits only after **every** worker thread has joined — [`src/core/worktree/list_stream.rs#L87-L96`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list_stream.rs#L87-L96).
5. Default columns (`ListColumn::list_defaults()`) don't include Size, and the default sort needs neither size nor mtime — the walk's output is never rendered.

The blocking path already does this correctly: `run_blocking` gates the size walk on `selected_columns.contains(&ListColumn::Size) || sort_spec.needs_size()` and mtime on `sort_spec.needs_mtime()` — [`src/commands/list.rs#L237-L238`](https://github.com/avihut/daft/blob/02af3209/src/commands/list.rs#L237-L238). Only the live path pays for hidden fields.

## Evidence

A/B under a PTY in the daft repo (5 worktrees, ~40 GB combined on disk — mostly `target/` dirs), identical visible output:

| run | real | user | sys |
|---|---|---|---|
| `daft list` (live path) | **2.07 s** | 0.33 s | **3.17 s** |
| `DAFT_NO_LIVE=1 daft list` | **0.45 s** | 0.15 s | 0.24 s |

(Warm FS cache; the reporter's interactive runs at 4 s reflect colder caches. `sys ≫ user` is the syscall storm.)

`sample` of the live process mid-hang: the main thread parked in `kevent` (renderer tick loop), one thread in `pthread_join` (waiting on collector workers), and the remaining worker threads inside a self-recursive frame (`walk`) whose leaves are all filesystem metadata syscalls. Top-of-stack totals across the 2 s sample window:

```
lstat               220   <- symlink_metadata per entry
__ulock_wait        158   <- join_thread waiting on workers
kevent              158   <- renderer loop, idle
semaphore_wait_trap 158   <- ctrl-c handler thread, idle
__open_nocancel      47   <- opendir
__getdirentries64    34   <- readdir
```

Per-worktree sizes the walk has to cover here (`du -sk`): 15.1 G, 14.8 G, 5.2 G, 4.2 G, 0.9 G.

## Why it feels like a recent regression

The mechanism shipped with the live-list infrastructure in #410 (v1.10.0, 2026-04-28) — `FieldSet::ALL` has been requested ever since. The tail grows with the combined file count of all worktrees and with FS cache temperature, so it creeps upward as `target/`/`node_modules/` accumulate and gets suddenly noticeable, without any recent code change. Scales per-project, matching the "30 s in other projects" observation.

## Suggested fix

In `run_live`, request only the fields actually needed instead of `FieldSet::ALL`:

- union of the fields backing `selected_columns` (a small `ListColumn → FieldSet` mapping — doesn't exist yet), plus
- `sort_spec.required_fields()` (already exists: [`src/core/sort.rs#L440`](https://github.com/avihut/daft/blob/02af3209/src/core/sort.rs#L440)), plus
- the line-stat fields only when `stat == Stat::Lines` (the workers already gate on this).

The collector already honors an arbitrary `FieldSet` (that's how post-fetch re-runs work), the loading-shimmer check only consults visible columns, and the completion sentinel fires when the narrower worker set joins — so this should be contained to `src/commands/list_live.rs` plus the mapping helper. The `FieldSet::ALL` rationale comment at `list_live.rs#L192-L198` needs updating in the same change.

Secondary observation (possibly separate scope): the collector spawns one unbounded thread per worktree ([`list_stream.rs#L112-L121`](https://github.com/avihut/daft/blob/02af3209/src/core/worktree/list_stream.rs#L112-L121)); when Size *is* legitimately requested, N concurrent full-tree walks contend on the disk.

## Acceptance criteria

- On a TTY with default columns/sort, `daft list` total wall time is in the same ballpark as `DAFT_NO_LIVE=1 daft list` — no full-tree walk (verifiable with `sample`: no `lstat` storm after cells land).
- `--columns +size` and `--sort -size`/`-activity` still populate correctly (walk runs only then).
- Regression test per CLAUDE.md: extract the requested-`FieldSet` computation into a pure function and unit-test that default columns + default sort exclude `SIZE` and `MTIME` (and that `+size`/size-sort include them).

## Workaround

`DAFT_NO_LIVE=1 daft list` — same table, blocking renderer, correctly gated fields.

avihut:
## Summary

`daft list` on a TTY rendered its table almost instantly, then hung for seconds before exiting (4 s in this repo, up to 30 s in `node_modules`-heavy projects). The live path requested `FieldSet::ALL` from the streaming collector, so every per-worktree worker also ran the hidden SIZE cluster (`compute_directory_size` — a full recursive walk of each worktree) plus MTIME, even though no default column or sort uses them — and the renderer can only exit once **all** workers join (`WorktreeInfoCollectionDone`).

The blocking path already gates these fields (`has_size` / `compute_mtime` in `run_blocking`); the live path was the only collector callsite passing `FieldSet::ALL`.

## The fix

- New `collector_fields(columns, sort_spec, stat)` helper in `src/commands/list_live.rs`: the union of what the selected columns render (mirrors the renderer's shimmer mapping) and what the sort inspects (`SortSpec::required_fields()`), upgraded with the `*_LINES` variants in `--stat lines` mode.
- `run_live` requests exactly that set, and passes the complement as `TuiState` seeded fields so never-streamed cells don't read as in-flight (loading shimmer, verbose footer counter).
- Exhaustive column match (no `_` arm) so a future `ListColumn` variant forces a decision here.

## Verification

- 7 new unit tests on `collector_fields` (defaults exclude SIZE/MTIME/`*_LINES`; `+size`, `--sort=-size`, `--sort=-activity`, and lines-mode each request what they need; a pure-seed `branch,path` view requests nothing).
- Pre-commit trio green (fmt, clippy, test:unit), 53/53 YAML list scenarios pass.
- E2E timing through a PTY harness that answers the cursor-position query (`ESC[6n`): 220k-file fixture 1.8–2.0 s → 0.23 s (≈ the `DAFT_NO_LIVE=1` baseline); real repo (6 worktrees, ~45 GB of build artifacts) 3.5 s cold / 1.1 s warm → ~0.3 s.
- Opt-in paths intact: `--columns +size` still walks and renders the Size column + TOTAL footer; `--sort=-size` / `--sort=-activity` collect what they need.
- Manual test plan: `test-plans/list-live-field-gating.md`.

Making the SIZE walk itself faster when requested is tracked separately in #668.

Fixes #665

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -575,19 +575,19 @@ Available in `run` commands:
 ### Skip and Only Conditions
 
 ```yaml
-skip: CI                         # Skip when env var is truthy
-skip: true                       # Always skip
+skip: CI # Skip when env var is truthy
+skip: true # Always skip
 skip:
-  - merge                        # Skip during merge
-  - rebase                       # Skip during rebase
-  - ref: "release/*"             # Skip if branch matches glob
-  - env: SKIP_HOOKS              # Skip if env var is truthy
-  - run: "test -f .skip-hooks"   # Skip if command exits 0
-    desc: "Skip file exists"     # Human-readable reason for the skip
+  - merge # Skip during merge
+  - rebase # Skip during rebase
+  - ref: "release/*" # Skip if branch matches glob
+  - env: SKIP_HOOKS # Skip if env var is truthy
+  - run: "test -f .skip-hooks" # Skip if command exits 0
+    desc: "Skip file exists" # Human-readable reason for the skip
 
 only:
-  - env: DEPLOY_ENABLED          # Only run when env var is set
-  - ref: "main"                  # Only run on main branch
+  - env: DEPLOY_ENABLED # Only run when env var is set
+  - ref: "main" # Only run on main branch
 ```
 
 ### Trust Management
--- a/src/commands/list_live.rs
+++ b/src/commands/list_live.rs
@@ -248,12 +248,11 @@ pub fn run_live(args: Args) -> Result<()> {
 }
 
 /// Fields the streaming collector must populate for this view: what the
-/// selected columns render plus what the sort inspects, with the `*_LINES`
-/// variants added in `--stat lines` mode (the renderer and `SortKey::compare`
-/// read those then; the summary bits stay in the set because the loading
-/// shimmer keys off them). The collector request routes through here so it
-/// can't drift back to `FieldSet::ALL`, whose hidden SIZE walk kept the
-/// renderer alive for seconds after the table was fully drawn (#665).
+/// selected columns render plus what the sort inspects
+/// (`SortSpec::required_fields`, itself `--stat`-aware). The collector
+/// request routes through here so it can't drift back to `FieldSet::ALL`,
+/// whose hidden SIZE walk kept the renderer alive for seconds after the
+/// table was fully drawn (#665).
 fn collector_fields(columns: &[ListColumn], sort_spec: Option<&SortSpec>, stat: Stat) -> FieldSet {
     let mut fields = FieldSet::EMPTY;
     for column in columns {
@@ -273,6 +272,10 @@ fn collector_fields(columns: &[ListColumn], sort_spec: Option<&SortSpec>, stat: 
         fields |= spec.required_fields();
     }
     if stat == Stat::Lines {
+        // Rendered diff columns display line counts in lines mode, so their
+        // `*_LINES` variants must be collected even without a matching sort
+        // key (sort keys arrive pre-upgraded from `required_fields`). The
+        // summary bits stay in the set — the loading shimmer keys off them.
         if fields.contains(FieldSet::BASE_AHEAD_BEHIND) {
             fields |= FieldSet::BASE_LINES;
         }
--- a/src/commands/list_live.rs
+++ b/src/commands/list_live.rs
@@ -12,7 +12,7 @@ use crate::{
         sort::SortSpec,
         worktree::{
             info_field::FieldSet,
-            list::{EntryKind, WorktreeInfo, collect_branch_info},
+            list::{EntryKind, Stat, WorktreeInfo, collect_branch_info},
             list_stream,
             sync_dag::{DagEvent, PatchSource},
         },
@@ -63,6 +63,12 @@ pub fn run_live(args: Args) -> Result<()> {
         ),
         None => None,
     };
+
+    // Request only the fields this view renders or sorts by. Requesting more
+    // (the old `FieldSet::ALL`) forced every worker through the SIZE cluster —
+    // a full recursive walk of each worktree — before the completion sentinel
+    // could fire and let the renderer exit (#665).
+    let fields = collector_fields(&selected_columns, sort_spec.as_ref(), stat);
 
     let show_local = args.branches || args.all;
     let show_remote = args.remotes || args.all;
@@ -170,7 +176,7 @@ pub fn run_live(args: Args) -> Result<()> {
     let collector_handle = list_stream::spawn(
         list_stream::CollectorRequest {
             targets,
-            fields: FieldSet::ALL,
+            fields,
             stat,
             source: PatchSource::Collector,
             ctx: collector_ctx,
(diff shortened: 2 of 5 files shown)


avihut:
Addressed all three review items in a8939ce8 + 5309c1e2:

**Medium — stat-blind `required_fields()`**: implemented as sketched — `required_fields()` now reports the `*_LINES` variants for Base/Changes/Remote when `self.stat == Stat::Lines` (added alongside the summary bits, since the Changes lines total also folds in the summary-cluster `untracked` count), with two new unit tests. One correction to the finding's caveat: the gap was not just latent. The line-stat clusters emit their own patches (`P::BaseLines` / `P::ChangesLines` / `P::RemoteLines` in cluster 7, after the summary clusters), so in lines mode a lines patch really did land with `touched ∩ required_fields() = ∅` and skipped the resort until the final CollectionDone re-sort. The `live_table.rs:113` callsite is fixed with no changes there — `run_live` already threads `.with_stat(stat)` into the spec.

The manual upgrade block in `collector_fields` stays, though: it covers *rendered* diff columns, which display line counts in lines mode even when no sort key touches them — `required_fields()` only speaks for the sort. Its comment now says exactly that.

**Low — stale `FieldSet::ALL` doc**: rewritten to describe its current role (fully-received sentinel for the inflight counter, outright seed for collector-less views like `repo remove`; the live list requests a narrowed set and seeds the complement).

**Observation — `!fields` relies on `ALL == u32::MAX` + bitwise `Not`**: converted from on-record to enforced — `info_field.rs` now has `complement_of_a_request_unions_back_to_all` pinning `requested | !requested == ALL` and `!EMPTY == ALL`, so either coupling breaking fails a test instead of a hallway conversation.

Full suite green locally (2039 passed / 0 failed).

Issue #478, opened 2026-05-09
fix(hooks): yaml_executor unit tests leak orphan UUID dirs into ~/.local/state/daft


avihut:
## Bug

Unit tests in `src/hooks/yaml_executor/mod.rs::tests` write into the user's real `~/.local/state/daft/jobs/` on every `cargo test` run, leaking orphan job records that accumulate forever.

## Background

Pre-#448, the leak was visible: `execute_yaml_hook_with_rc` used the cwd-based `compute_repo_id()` (rather than `compute_repo_id_from_common_dir(&ctx.git_dir)`), so unit tests run from this very repo silently picked up the real repo's `daft-id` UUID and wrote invocation records under the live worktree name `feature/new` (hardcoded in `make_ctx_with_dir`). That's the giant `feature/new` mountain visible in `daft hooks jobs --all`.

#448 fixed the cwd issue by switching to `ctx.git_dir` (yaml_executor/mod.rs:158), so tests no longer pollute the *current* repo's UUID dir. **But the leak isn't fully gone**: `LogStore::for_repo(repo_hash)` still resolves to `daft_state_dir().join("jobs").join(repo_hash)` (log_store.rs:140), and `daft_state_dir()` always returns `~/.local/state/daft` outside of dev-build env-overrides. Each test run now creates a *fresh orphan repo-UUID directory* there — invisible from any real repo's `daft hooks jobs --all`, but accumulating forever.

This is also state pollution that can affect parallel test runs interacting with shared filesystem state, and risks slow filesystem operations as the orphan dir count grows.

## Repro

```sh
ls ~/.local/state/daft/jobs | wc -l
mise run test:unit  # runs yaml_executor tests
ls ~/.local/state/daft/jobs | wc -l   # grows by ~25 fresh orphan UUIDs per run
```

## Proposed fix

Inject the state dir base into the LogStore path:

1. Add `LogStore::for_repo_in(repo_hash, state_base)` in `src/coordinator/log_store.rs`.
2. Make existing `LogStore::for_repo(repo_hash)` delegate via `daft_state_dir()`.
3. Add an optional `state_dir: Option<PathBuf>` field on `HookContext` (with a `with_state_dir(p)` builder).
4. In `execute_yaml_hook_with_rc`, use `LogStore::for_repo_in(&repo_hash, &ctx.state_dir.clone().unwrap_or_else(daft_state_dir)?)`.
[...]

PR #698, opened 2026-07-12
test: real-state tripwire + fix repo-catalog/state test-isolation leaks

### LINKED ISSUE #697: Repo-catalog integration tests leak into the real `catalog.db` — clean up and enforce test isolation
## Symptom

Repo-catalog integration tests wrote throwaway test entries into the **real** global
catalog store — `catalog/catalog.db` under `daft_data_dir()` (on macOS,
`~/Library/Application Support/daft/catalog/catalog.db`). The catalog is the first
consumer of the `data_dir` XDG surface, and test-origin rows ended up in the
production catalog on a real machine.

Two deliverables, one ticket:

1. **Clean up** the leaked test rows from the real `catalog.db` (one-off manual op — see below).
2. **Fix the tests** so catalog operations can never write the real data dir again, with a tripwire that makes a regression impossible to land silently.

## Root cause — investigation

The catalog DB path routes through `daft_data_dir()`:

- `src/store/paths.rs:75-86` — `catalog_db()` → `catalog_db_under(&daft_data_dir()?)`, i.e. `<data_dir>/catalog/catalog.db`.
- `src/lib.rs:81-89` — `daft_data_dir()` honors `DAFT_DATA_DIR` **only** under `cfg!(any(daft_dev_build, test))`; otherwise it resolves the real XDG data dir.

Both test harnesses **already export `DAFT_DATA_DIR`** — so this is *not* a missing-isolation-var:

- YAML runner — `xtask/src/manual_test/daft_executor.rs:99, 154-155` (per-sandbox `daft-data`).
- Bash suite — `tests/integration/test_framework.sh:136-137` (`$TEMP_BASE_DIR/ddata`).

Therefore the binary under test **ignored the exported `DAFT_DATA_DIR`**, which means it was **not a `daft_dev_build`**. Two known ways that happens:

- **(a) #669 mechanism** — a build at a tagged-release commit drops `daft_dev_build`, compiling all `DAFT_*_DIR` overrides out; every daft invocation in the post-release window writes real dirs. #669 tracks this for the `state_dir` surface (pre-push / coordinator); the catalog adds a **new `data_dir` surface with the same failure**.
- **(b) Binary selection** — the harness (or a manual `mise run test:manual` invocation) ran an installed/release `daft` instead of the dev build.

Less likely but worth ruling out during the fix: a catalog write path that bypasses `daft_data_dir()` (check `catalog_db_probe()` at `src/store/paths.rs:111`, and any detached background process that opens the catalog).

## Cleanup (one-off, on the real machine — back up first)

- Locate the real catalog: `catalog/catalog.db` under the data dir (macOS: `~/Library/Application Support/daft/catalog/catalog.db`).
- **Copy the file aside first.**
- Inspect catalog rows for entries whose repo path points at throwaway test locations — mirror the #666 forensic signature (`/private/tmp/...`, `/tmp/daft-*`, `git-worktree-integration-tests`, sandbox `daft-data` bases).
- Delete only the test-origin rows (or reset the catalog if it holds nothing but test entries). Verify real repo entries survive.

## Fix directions

- **Tripwire (primary):** snapshot mtime/hash of the real `catalog/` dir (alongside `config` + `state`) before each suite run and fail if it changed — locally and in CI. #667 proposes this for config/state; extend it to cover the `data_dir`/catalog surface. "No writes to the real catalog" becomes an explicit, enforced acceptance criterion.
- **Guard the dev-build assumption:** the suites must run a `daft_dev_build` binary and fail loudly if the selected binary ignores `DAFT_*_DIR` (closes the #669 release-window hole for the catalog surface, not just state).
- If a catalog path resolves without the override, route it through `daft_data_dir()`.
- **Regression test** (per the bug-fix rule): catalog operations under a sandboxed `DAFT_DATA_DIR` must leave the real catalog dir untouched (a YAML scenario under `tests/manual/scenarios/repo/` plus/or a store-level unit test).

## Relationships

- **#669** — `daft_dev_build` drop at release tags; same root mechanism, `state_dir` surface. This ticket is the `data_dir`/catalog instance and the missing tripwire.
- **#667** — bash-suite isolation + the tripwire idea, `config_dir` surface.
- **#666** — `repos.json` corruption from the same isolation-leak class.
- **#478** — unit-test state-dir leak (prior art, same class).
- **#447** — retire the bash suite; "no writes to real user state (incl. catalog)" should be an explicit acceptance criterion there too.

avihut:
## What

The repo-catalog integration suites leaked throwaway test repos into the
developer's **real** `catalog.db`. Two deliverables, per the ticket:

**1. Cleanup (manual, done on the dev machine).** The real catalog had
accumulated 823 rows — 822 test-origin
(`/private/tmp/git-worktree-integration-tests/...`, `/var/folders` sandboxes)
and exactly 1 real (`daft` itself). Backed the file up, deleted the 822
test rows, kept the real one; `application_id`/`user_version` header pragmas and
WAL left intact.

**2. Enforcement (this PR).** Make "no writes to the real catalog / state /
config" an enforced, un-regressable invariant.

## The tripwire

`xtask real-state-guard {snapshot,verify}` fingerprints the real
config/state/data dirs before each suite and fails the run if they change —
locally and in CI — wired into `test:{unit,integration,manual}` and `ci` through
`mise-tasks/test/_state_guard_lib.sh` (`DAFT_SKIP_STATE_GUARD=1` escape hatch).

- **Targeted coverage**, not a whole-XDG-dir walk — on macOS the data dir also
  hosts centralized-layout worktrees: content-hash the `<data>/daft/catalog/`
  DB triplet and `<config>/daft/repos.json`; compact entry-set digest for the
  litter-prone `<state>/daft/` + `jobs/`.
- Resolves the real dirs via `dirs` (never reads `DAFT_*_DIR`, so it always
  targets the real surface), never creates them; a drift test pins its
  resolution to daft's own `daft_{config,data,state}_dir()`.
- A read-only **`daft __dirs` preflight** fails fast if the binary under test
  ignores `DAFT_*_DIR` — a non-dev-build, or a system `daft` on `PATH` —
  closing the #669 release-window hole for the data surface *before* it can
  leak.

## Leaks the tripwire surfaced (and this PR fixes)

- A store-level regression test locking `catalog_db()` / `catalog_db_probe()` to
  the sandboxed `DAFT_DATA_DIR`.
- Turning the tripwire on revealed a **pre-existing** state-dir leak (#478/#669
  class): 9 unit tests in `branch_delete` + `install` open the coordinator
  store via `paths::for_repo` while running `worktree-post-remove` /
  visitor-seed code, writing the real `~/.local/state/daft/jobs/<uuid>/`. Fixed
  with a shared `#[serial] store::paths::IsolatedStateDir` guard. Backtrace
  tracing confirmed those were the only leak sites; the guarded `test:unit` is
  now green.

## Testing

- `mise run fmt` / `clippy` / `test:unit` green (`test:unit` now runs under the
  tripwire; pre-push hook passed).
- Guard pass/fail paths validated end-to-end: the preflight accepts a dev binary
  and rejects an ignore-overrides binary; the tripwire passes a clean run and
  fires on a simulated real-catalog write.

Fixes #697.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -101,6 +101,20 @@ a substitute. Production two-stage probes that expect
 `fatal: not a git repository` as a negative signal must suppress the stderr at
 the call site.
 
+**Never let a test write daft's real config/state/data dirs.** A tripwire —
+`xtask real-state-guard`, wired into `test:{unit,integration,manual}` and `ci`
+through `mise-tasks/test/_state_guard_lib.sh` — fingerprints the real
+`~/Library/Application Support/daft` (catalog DB + `repos.json`) and
+`~/.local/state/daft` (`jobs/` entry set) before each suite and fails the run if
+they change, locally and in CI. Tests that exercise hook / visitor-seed code
+paths open the coordinator store via `paths::for_repo`, which resolves
+`daft_state_dir()`; wrap them in `#[serial]` +
+`store::paths::IsolatedStateDir::new()` so they write a tempdir, not the
+developer's real state. Suites that exec a built binary also run a read-only
+`daft __dirs` preflight that fails fast if the binary ignores `DAFT_*_DIR` (a
+release/tagged build, or a system `daft` on `PATH`). See #697; prior art
+#478/#669.
+
 ## XDG Conventions
 
 This project follows the XDG Base Directory Specification. Use the `dirs` crate
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5687,6 +5687,7 @@ dependencies = [
  "crossterm",
  "ctrlc",
  "daft",
+ "dirs 6.0.0",
  "indicatif",
  "nix",
  "rayon",
--- a/mise-tasks/test/_state_guard_lib.sh
+++ b/mise-tasks/test/_state_guard_lib.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# Sourceable helper: the real-state test-isolation tripwire (#697).
+#
+# Guards against the leak class where a test run writes daft's *real*
+# config/state/data dirs instead of its sandbox — the failure that put 822
+# throwaway /tmp repos into the developer's real catalog.db. Two layers:
+#
+#   * assert_binary_honors_overrides <daft-bin> — a READ-ONLY preflight. Runs
+#     `<daft-bin> __dirs` under a throwaway DAFT_*_DIR sandbox and fails if the
+#     binary resolves the real dirs instead. Catches a non-dev-build binary (a
+#     release/tagged build with the DAFT_*_DIR overrides compiled out) or a
+#     system `daft` shadowing the local one on PATH — *before* the suite can
+#     leak. Only meaningful for suites that exec a built binary (integration,
+#     manual); unit tests honor the overrides by construction under cfg(test).
+#
+#   * with_state_guard <cmd...> — snapshots the real dirs, runs the command,
+#     then re-verifies them via `xtask real-state-guard`. A change is fatal
+#     (exit 1) regardless of the command's own exit code; an unchanged run
+#     passes the command's exit code through so genuine test failures still
+#     surface.
+#
+# Escape hatch: DAFT_SKIP_STATE_GUARD=1 disables both layers.
+#
+# File is intentionally non-executable so mise hides it from `mise tasks ls`.
+
+# Invoke the xtask guard. Centralized so the snapshot and verify calls stay in
+# lockstep (same package, same quieting).
+_state_guard_xtask() {
+  cargo run -q --package xtask -- real-state-guard "$@"
+}
+
+# Read-only preflight: assert the daft binary at $1 honors DAFT_*_DIR.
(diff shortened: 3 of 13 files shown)


avihut:
Thanks for the thorough review. Confirming the two items you flagged as unverifiable from the sandbox:

- **CI is green** — all 15 checks pass, including `integration-tests (default/gitoxide × bash/yaml)`, which now run *under* the tripwire. So the guard wiring + `__dirs` preflight are validated in CI against the real suites, not just locally.
- **Coverage empirically confirmed** — the full `test:unit` (locally ×2, the pre-push hook, and CI) runs under the tripwire and is green, so any missed `SeedsContext::open` leak site would have tripped it. That matches your static trace (the seed-store open is gated behind `!files_propagated.is_empty()`, and the `checkout_branch` fixtures use a *tracked* `daft.yml` that never goes through visitor propagation): 9 sites, all fixed.

The FNV-1a-64 choice (#1) and the `exit 1`-from-a-sourced-function (#2) are both intentional as you noted — leaving them as-is.

Issue #387, opened 2026-04-25
feat: copy_paths in daft.yml for replicating gitignored paths between worktrees


avihut:
## Goal

Let users declare a list of gitignored paths (`target/`, `node_modules/`, `.venv/`) that daft replicates between worktrees on creation, using reflink/CoW where the filesystem supports it. The list lives in `daft.yml` as a `copy_paths:` block, not a separate file.

## Surface

```yaml
# daft.yml
copy_paths:
  - target/                  # Rust build cache
  - node_modules/            # Node deps; safe to copy when bitwise-identical
  - .venv/                   # Python venv; comment-friendly per entry
  - "**/dist/"               # glob support
copy_paths_options:
  reflink: prefer            # prefer | only | never
  max_size: 5GB              # skip individual paths exceeding this
  fail_on_skip: false        # if true, surface a warning instead of silent skip
```

- Built-in step: `daft worktree-copy-ignored` callable directly and from `worktree-post-create` hooks
- Hook composition: a `worktree-post-create` job can reference the same list (OS-gating, `tracks`, etc.) for users who want full hook-level control instead of the built-in step

## Why daft.yml instead of a `.daftinclude` file

- daft.yml is already the single source of truth for hooks, layouts, and (per #335) visitor configuration. A separate file duplicates the surface.
- YAML comments on each entry give the per-line documentation appeal of a flat file.
- Visitor configuration (#335) extends naturally: users can override `copy_paths` in their local-only daft.yml without inventing a `.daftinclude.local` parallel.
- Hook integration is cleaner: the same engine that runs `worktree-post-create` can invoke the copy step, with consistent template variables and OS gating.

## Migration note for worktrunk users

Document explicitly: the equivalent of worktrunk's `.worktreeinclude` is `copy_paths:` in `daft.yml`. Same semantics, single config file. Include a one-liner in the docs that converts `.worktreeinclude` to a `copy_paths` block so migrators don't have to think.

## Distinct from existing primitives

- `git-worktree-carry` moves *uncommitted working-tree changes* from one worktree to another. WIP transfer, not build-cache replication.
- Imperative `worktree-post-create` hooks (`run: cp -a ...`) work but force every user to reinvent reflink fallback, size thresholds, and gitignore validation. The declarative form centralizes this.

## Implementation

- Use `reflink-copy` crate (ergonomic Rust wrapper)
[...]

PR #796, opened 2026-07-29
feat(config): a settings interface that shows where every value comes from

### LINKED ISSUE #470: Config management interface
daft configurations have grown very big and complex and it's become hard to know what to set when, and what is even configurable. Create a TUI interface for aggregating all of the configurations and set them for local and global configs.

avihut:
Fixes #470

daft has ~180 knobs across six storage backends resolved through five different
precedence chains, and no way to answer the two questions people actually
arrive with: *what is my configuration*, and *where did this value come from*.
`git config --list` shows the keys daft happens to read but not the ones it
would read; the docs describe defaults that have drifted from the code; and
nothing anywhere shows that a local value lost to a global one, or that a value
is set and inert.

This lands a full-screen browser over every setting daft has, a scriptable CLI
underneath it, and — the part that makes both possible — a machine-readable
registry with a drift gate holding it complete.

## What lands

**A settings registry** (`src/core/settings_spec.rs`). One row per setting:
key, label, help, category, value type, default, backend, and the odd
properties that were previously only expressible as loader behaviour —
`global_only`, `env_override`, `deprecated_alias`, `inherits`, `managed_by`,
plus a cross-key `validate` hook. 88 rows: **77 git-config keys** (63 fixed + 14
per-hook), **10 `daft.yml` scalars**, and **the worktree layout** as one row
with its six-layer chain.

Two conventions keep the rows honest rather than merely present: they reference
the existing `keys::` consts, so a rename is a compile error and not an orphan;
and enum value sets come from each parsing type's new `variants()`, declared
next to its `parse()`, so the registry and the parser cannot disagree about what
a setting accepts.

**A drift gate** (`config_registry_drift` in `xtask`). Scans `src/**/*.rs` for
key-shaped `"daft.*"` literals and fails on any the registry does not claim —
and runs the converse, so a row for a key nothing reads fails too. It rides the
existing `xtask-test` CI job, the `xtask-tests` pre-merge ring, and
`mise run test:xtask`: **zero new parity surfaces**. It compares through git's
own case rules, so `daft.checkout.PUSHVERIFY` passes and a mis-cased
*subsection* — a different key to git, and inert — does not.

It earned its place during development by catching three of my own test
fixtures.

**The screen** (`daft config`, `src/commands/config/screen/`). Settings grouped
by what they do, with storage as metadata rather than structure. The detail
panel is a provenance ladder: every layer's answer, a ● on the one daft reads,
and the file it came from. `enter` opens a modal editor, `space` flips a
boolean, `u` unsets, `s` moves the write-scope pill, `/` filters across label,
key and help together, `]`/`[` walk the categories. Writes are immediate,
narrated, and re-resolve everything — a write can move which layer wins for
*other* settings, and a stale ladder is the one thing this screen cannot show.

**The CLI** — `list`, `get`, `set`, `unset`. `get` keeps `git config --get`'s
contract (exit 1 when unset, value on stdout, nothing else) so it drops into
scripts; `--origin` prints the ladder; `list --format json` carries the whole
read model. Every write goes through one path shared with the screen, so the
type check, the cross-key merge rule, and the tightening rule cannot be reached
around.

**Diagnostics as UI.** A value that will not parse is red in the row, in the
ladder, and in an explanation; a deprecated alias names its replacement; a
global-only key set at local scope says it is never read; an env-shadowed value
says no config file can change it. `daft.*` keys that match no registry row get
their own rows under an Issues rail entry with a did-you-mean — the usual cause
is a mis-cased subsection, where the entire difference is one letter's case.

## Worth a reviewer's attention

**The resolver reproduces the loaders' rules rather than tidier ones.** A
`global_only` key's local value is inert and says so. A deprecated alias is a
whole-ladder fallback, not a per-scope one. And an unparseable value falls back
to the **default**, not to the next scope down — so a broken local value masks a
good global one, and the ladder shows exactly that. Each of these is a rule some
loader already had; the screen would be lying if it smoothed them over.

**Two write paths deliberately have no unit test.** Setting the layout or a
`daft.yml` scalar touches real state: the global TOML, the trust DB, or a
`daft.local.yml` in whatever worktree the test runs in. I made both mistakes
during development and reverted both; the code says why, and those paths are
covered by sandboxed YAML scenarios instead.

**The `daft.yml` editor is line surgery, never a serializer round-trip** —
`daft.yml` is tracked, so a reflow would land in the user's diff. Its safety
property is the post-edit check: re-parse as a real `YamlConfig` *and* compare
whole documents with the target path removed, because an editor that drops a
sibling key still parses. Exotic shapes (flow style, anchors, duplicate keys)
bail to "edit it by hand" rather than risk a mangled committed file.

**`git config` writers now route through `crate::utils::git_command_at`**, so an
inherited `GIT_DIR` — daft running inside a git hook — cannot retarget a write
at the parent repository. Same hazard the read path already documented.

## Coordination note

Two branches in flight add settings keys (#468 hooks, #387 copy). Once this
lands, their PRs fail `test:xtask` until they add a registry row each. That
friction is the point: it is a one-time build-time check replacing an unbounded
per-future-PR coordination tax, and it is the same shape as the
`.github/labeler.yml` glob rule and the `src/store/repos/` format-gate.

## Verification

- `mise run clippy` clean; `mise run test:unit` 3400 passed
- `cargo test -p xtask` 239 passed (includes the drift gate)
- `mise run test:integration` full matrix, and the new
  `test_config.sh` (16) + `test_config_tui.sh` (3, driven under a real PTY)
- `mise run test:manual tests/manual/scenarios/config` — 12 scenarios, 66 steps
- `mise run completions:test` 34 passed; `man:verify` and `docs:cli:verify` clean
- `test-plans/config-management.md` carries the manual checklist

## Not in this PR

`config remote-sync` keeps its own 670-line TUI for now. Repointing it at the
new screen and reimplementing `--on/--off/--status` over the shared write path
is the one plan item left, and it is self-contained enough to land separately.

Follow-ups worth issues: refactoring `DaftSettings::load` onto the registry
(and closing the 12-key gap where `load_global` reads fewer keys than
`load_with`); the four dead `daft.yml` keys (`colors`, `no_tty`, `output`,
`source_dir_local`) that are parsed and never read; generating the
`docs/reference/configuration.md` tables from the registry; editing structured
`daft.yml` documents (`hooks:`, `tasks:`, `relations:`), which keep their own
commands; and `shared:` list editing, which belongs to `daft shared`.


avihut:
@avihut — high-effort review run over the branch: **10 findings, all confirmed by an independent verify pass, all fixed** in six commits (`2cda483d..8d8eb149`). Every check on the PR is green.

Two structural themes, not ten unrelated bugs.

## The `daft.yml` scope model was wrong in four ways

The 77 git-config rows were sound. The 10 `daft.yml` rows were not, and every fault pointed the same direction — the screen said one thing, the loader did another:

- **A local write landed in a file nothing reads.** In a repo with no `daft.yml` — the common case, it's opt-in — `daft config set log.retention 30d` wrote `daft.local.yml` and reported success. The loader derives the overlay's name from the main config and gives up before merging when there is none, so the value never took effect and `get` kept printing the default. Refused now, at the one seam both the CLI and the modal pass through.
- **`--global` wrote the repository's committed `daft.yml`** while narrating and documenting it as user-wide. Someone setting a personal preference edited the team's file and found it in `git status`. Neither file is user-wide, so neither borrows git's words any more: they're "committed config" and "local overlay", extending the special case the layout row's two stores already had.
- **`unset` skipped the tighten-only check `set` enforces.** `set merge.ff any` was correctly refused; `unset merge.ff` deleted the gate outright and printed success. Same hole, one keystroke cheaper.
- **The read path ignored `extends:`.** A repo that factors its config out was told `log.retention` was on its 7d default while the daemon pruned at 30d, with an empty ladder to explain it. That's the one failure a provenance browser cannot have. The chain is walked as separate attributed documents rather than taken from `load_merged_config` — a merged config can't say which file a value came from, which is the only question the ladder answers.

With the chain visible the editor could finally start on it: it pre-filled from a git-rung lookup, so **every** `daft.yml` and layout row opened blank, contradicting the contract that Enter without moving is a no-op.

## The rest

- **`git config --unset` spends exit 5 on two opposite outcomes** — "not set", and "several values, I won't guess". Reading it as only the first printed "was not set" about a value still in force, while `config_set` errored loudly on the very same key. Disambiguated by asking git, not by matching its stderr — that text isn't an interface.
- **The tracked-file editor mangled comments two ways**: a `#` inside a quoted value was taken as a comment start (splicing the old value's tail into the file), and a new child key was planted under the *next* section's heading. Both passed `verify`, which compares parsed documents and structurally cannot see a comment — which is why the new tests assert exact bytes, not parsed values.
- **`daft conf<TAB>` stopped completing** in bash and zsh: the config block keys off the preceding word being `config`, already true while that word is being typed. fish gates on an already-seen word and was never affected.
- **The list froze on short terminals.** Scroll height was derived by subtracting a fixed chrome total while the layout gives the list `Min(3)` — the solver squeezes the detail pane instead. At 16 rows the subtraction hit zero and `follow_cursor` early-returned; `j` left a frozen list while the cursor walked off screen.

## Verification

24 new regression tests, each failing before its fix. Every gate green locally and in CI: unit 3424, xtask 239 (drift gate), 13 config scenarios / 74 steps, `test_config.sh` 16/16, `test_config_tui.sh` 3/3 under a real PTY, completions 35, clippy clean — plus the full bash integration matrix, all four cells.

Three of the fixes add a **refusal**, and `fmt`/`clippy`/`test:unit` don't run the YAML or bash suites — so I checked the existing scenarios for anything asserting the old success before committing, and ran both suites after. Nothing regressed.

## Worth your judgement, not mine

The tests for the three write-path fixes are scenarios rather than unit tests, for the reason already in the PR body: those paths touch real state. The refusals themselves are unit-tested, because they bail before any I/O.

And one thing this review makes sharper — `config remote-sync` isn't just unfinished, it doesn't map onto the new screen. It's a preset over three keys that live in two different categories (`checkout.fetch` and `checkout.push` under Checkout, `branchDelete.remote` under Push & Sync). So the plan's "deep-link into Push & Sync" would land you on a category holding one of its three keys. Whatever replaces it wants to be a filter across the three, not a category jump — worth deciding before that work starts.

avihut:
@avihut — amending the comment above. A second pass over my own fixes found three gaps, now closed in `5594cb84`. **All 17 checks green again, `mergeStateStatus=CLEAN`.**

Two of them mattered, and both were cases where the fix looked right because the tests couldn't see the shape that breaks it.

**The editor fix was wrong in visitor repositories.** I told the two `daft.yml` write targets apart by `tracked` — but `tracked` isn't a property of *the file*, it says whether the **main** config is committed. In a visitor's repo that's false for the main file and the overlay alike, so Local matched the main file (pushed first) and Global matched nothing and pre-filled blank — the exact symptom the change was meant to remove, relocated to a different repo shape. Every test I wrote built main as `tracked: true`, so the suite structurally could not catch it. `Layer::Yaml` now carries `overlay` alongside `tracked`; they're independent, and only one of them answers "which file does a write go to".

**The scope labels were fixed at the narration and nowhere else.** The finding said "narrated *and documented*" and I only did the first half:

- The header pill rendered `WriteScope::label()`, so the screen still said **"global"** over a write bound for the repository's committed `daft.yml` — the exact sentence the labels were changed to stop saying. It reads the row under the cursor now. (The modal's scope row was already correct.)
- The `--global` help said "global config instead of this repository", false in both clauses for a `daft.yml` key. It now names the three stores it can mean and says plainly that one of them is tracked. Man pages and CLI docs regenerated in the same commit.

Third, minor: `still_set`'s body didn't match its own comment — it read an empty pipe as "not set" whatever the exit status, putting a small silent success back into the function written to remove one. Practically unreachable, since exit 5 proves git already read the config, but gated on exit 1 now.

Re-verified after: unit 3416, xtask 239, 13 config scenarios / 74 steps, `test_config.sh` 16/16, `test_config_tui.sh` 3/3 PTY, clippy clean, `man:verify` + `docs:cli:verify` clean, and the full integration matrix in CI.

Worth saying plainly, since it bears on how you read the comment above: **my own verification passed on all three of these.** What caught them was re-reading the fixes against repository shapes the tests didn't cover — a visitor repo, and the difference between where a value is *narrated* and where it is *rendered*. If you review one thing here, make it `value_written_at` and the header pill.

PR #797, opened 2026-07-29
feat: copy build caches into new worktrees (copy: and daft warm)

### LINKED ISSUE #387: feat: copy_paths in daft.yml for replicating gitignored paths between worktrees

avihut:
Declarative CoW replication of gitignored build caches into new worktrees, plus
`daft warm` to replay it on demand.

## The surface

```yaml
copy: # short form
  - target/
  - node_modules/
  - "**/dist/"
```

```yaml
copy: # full form
  paths: [target/, node_modules/]
  fallback: copy # copy | skip — when the filesystem cannot reflink
  max_size: 5GB # per-ENTRY cap; gates the byte-copy fallback only
```

The stage runs between `shared:` symlinking and `worktree-post-create`, so a
hook-driven `npm install` / `cargo build` reconciles a warm tree instead of
building one. Wired into all three creation journeys that have a source
worktree — `daft go`, `daft start`, and sandboxes/forks. `daft clone` does not
run it: a fresh clone has nothing to copy from.

Reflink capability is decided by *attempting* one, never by sniffing the
filesystem, because a single entry can straddle mount points. Entries must be
gitignored — `check-ignore` **and** nothing tracked underneath, which catches
the force-added file inside an otherwise-ignored directory.

**It never aborts creation.** Every failure is a warning row; the worktree you
asked for still arrives. One rail row per declared entry, dim `○` for the stage
working as designed and yellow `↓` for a declaration daft could not honor —
fourteen distinct skip reasons, and no copy row is ever red.

## `daft warm`

```bash
daft warm              # warm the current worktree
daft warm feature-x    # warm that one
daft warm --from main  # name the source outright
daft warm --force      # replace entries already present in the target
```

`--force` is the only thing here that deletes, so it is the only thing that
interrogates the target first: content the **target tracks** is never replaced,
the bare container is refused as either end, and nothing is removed until the
copy is actually going to proceed — so `--force` can never report a skip over a
cache it already destroyed.

## Where the caches come from

The source is **ranked against what the new worktree will hold**, not taken
from the current directory:

1. the base branch's own worktree — `daft start feature-b master` copies
   master's caches from wherever you typed it;
2. any worktree sitting at the identical commit — sandboxes and forks included,
   since content identity is the whole criterion;
3. the worktree you ran from, as the floor.

Ties on rank 2 go to where you are standing, then to the warmest (which
declared entries exist, and how recently they were touched). The rail's section
anchor names the winner — `copied paths from 'master' · same commit` — because
the source is a ranked decision rather than something the command line shows.
`daft warm` runs the same ladder anchored on the target's own commit; `--from`
is never second-guessed.

The **configuration** still comes from the worktree you are standing in, so a
visitor's untracked `daft.local.yml` declaring `copy:` keeps working. Config
comes from where you are; bytes come from what matches.

## Also here

`daft doctor` reports the repository filesystem's reflink capability, and this
repo now declares its own `copy:` block (`target/`, `node_modules/`,
`docs/node_modules/` with `fallback: skip` — 18 GB is free on APFS and is not
free anywhere else).

## Verification

fmt · clippy · 3394 unit tests · man/CLI-doc verify · prettier · docs site
build · 22 copy scenarios (125 steps) · `test_warm.sh`.

Docs: `copy:` in the daft.yml reference, a "Copying build caches into new
worktrees" how-to with per-toolchain expectations, the generated `daft warm`
CLI page and man pages, the progress-timeline reference, and SKILL.md.

Fixes #387


avihut:
@avihut — xhigh multi-agent review round is **done and green**. Head `6854dc75`, **16/16 checks passing** (`dep-age-check` skipping, as designed for a branch with no dependency changes).

The review ran 6 finders → 51 candidates → 48 independent verifiers → **15 confirmed defects**, 5 refuted. All 15 are fixed across 6 commits, each with regression coverage.

## The three that could have destroyed or corrupted something

**`--force` deleted outside the worktree.** `target.join(rel)` is a lexical join and the kernel is not, so a symlinked directory on the way redirected both the write and the removal. That isn't hypothetical — it's what `shared:` installs: `shared: [web]` + `copy: [web/dist]` made every worktree's `web/dist` the *same* directory in daft's store, written into by a plain run and deleted out of by `--force`, green row either way. Now refused on both paths.

**Workspace symlinks were re-pointed at the source worktree.** The rebasing boundary was the copied *entry*, but under `copy:` the whole worktree moves. So `node_modules/@acme/api -> ../../packages/api` — every npm/pnpm monorepo — was rewritten to point back into the branch you were seeded from: the new worktree compiled the wrong sources, and every link dangled the moment that worktree was removed.

**`shared:` linking silently defeated the copy.** Linking runs `create_dir_all` on the parents it needs, so `shared: [node_modules/.cache/token]` manufactured an empty `node_modules/` that the copy stage then reported as `already present` — the cache never copied, in any worktree, on creation or on any later `daft warm`, with a dim row saying everything was fine.

## Three decisions worth your eye, because they changed behaviour you specified

1. **Stage order flipped — caches now copy *before* `shared:` linking**, in all three journeys. That's the fix for the masking bug above, and it's the right precedence anyway (`copy:` moves bulk content, `shared:` owns the specific paths it names, so links go on top). Visible change: the rail lists `copied paths` above `shared files` now.

2. **A cold same-commit worktree no longer wins the ladder.** The review found that warmth was consulted for two-or-more candidates but skipped entirely for a sole candidate — so the cold sandbox `daft go <tag>` leaves behind beat a fully-built source and copied nothing. Making warmth a *gate* would have contradicted the ladder you locked, so it only demotes when the rank-2 match carries **none** of the declared caches **and** something warmer exists. A repo where nothing is built yet still resolves exactly as you specified, and the rail says why (`· the same-commit worktree is empty`). Stale worktrees `git worktree list` still reports after deletion are now ineligible too.

3. **`--force` now replaces a wrong-shape destination** — previously the shape check returned before the force branch was read, so the one state no other run could repair was permanently stuck. The exception is narrow: shape-*changing* replacement of a link resolving outside the worktree. Replacing a link with a link stays allowed — an existing test encodes that, and it caught my first guard being too broad.

## The rest

Tag-over-branch anchor resolution (`daft start hotfix release` with both a `release` branch and tag took the tag's OID); the `origin` hardcoded in `warm`'s catalog rung, ignoring `daft.remote`; `warm` from a container root aborting instead of falling back to the default branch; `fallback: skip` walking every cache for a byte count nothing read (this repo's own `daft.yml` is that config); glob expansion descending into nested worktrees and sibling caches; no `Started` event so the rail showed nothing for the whole copy; a misspelled `paths:` key producing complete silence; and `warm --force` deleting the directory you're standing in with no cd-rescue.

`CopyConfig`'s untagged derive is now a hand-written `Deserialize`: `copy: target/` and `paths: target/` work, and a bad `fallback:` names the value and the alternatives instead of "data did not match any variant of untagged enum CopyConfig".

## Verification

New coverage: 7 engine tests, 3 ladder tests, 1 `cow_copy` boundary test, 1 `warm` container-root test, 1 wrapper integration test, and a YAML scenario for the `shared:`/`copy:` overlap — which I verified **fails with the old ordering** (2 of its 6 steps) and passes with the fix. Local: fmt, clippy, 3408 unit tests, man/CLI verify, 23 copy scenarios / 131 steps.

## Two things I did not fix, deliberately

- **`tests/integration/test_shell_init.sh` is not in `test_all.sh`, so CI never runs it.** CLAUDE.md designates it as the required home for shell-wrapper coverage, so the `warm --force` cd-rescue regression lives there and passes locally — but it does not gate this PR. Two of its assertions have also been red on master all along: they grep for `alias gwcob=` where `gwcob` is emitted as a shell *function*. Both want an issue, not a widened PR.
- **Plain/piped mode still prints no source provenance** — the rail carries it, the non-TTY fallback doesn't. Flagged earlier, still unbuilt.

One self-inflicted red along the way: my first test for the measurement-gate split observed a `-v` line that `fallback: skip` only emits on a reflinking filesystem, so it passed on APFS and failed on Linux CI. Fixed by asserting the named predicates instead (`6854dc75`).

